PR #56, opened 2018-03-20
added empty clue checkbox

mikellama:


rogers-j:
I think the difference cannot just be a checkbox. I think it would be more accurate to say that games with this setting are a _different variant_ entirely.

mikellama:
What about a checkbox that toggles between normal and empty-clue variant groups? Or should we start with just no variant and rainbow options added to the bottom of the variant list?

Zamiell:
just have a checkbox

PR #60, opened 2018-06-22
Adding SingleMultiRainbow variant and an option to reverse hands.

jeep:


public/js/ui.js
```diff
@@ -1232,7 +1232,7 @@ function HanabiUI(lobby, gameID) {
         Kinetic.Group.call(this, config);
 
         this.align = (config.align || 'left');
-        this.reverse = (config.reverse || false);
+        this.reverse = ( lobby.reverseHands ? (config.reverse || false) : !config.reverse);
```
rogers-j: You should check the lobby setting when `config.reverse` gets its value, no?

public/js/constants.js
```diff
@@ -938,6 +952,7 @@
         exports.VARIANT.AMBIGUOUS,
         exports.VARIANT.BLUERED,
         exports.VARIANT.ACIDTRIP,
+	exports.VARIANT.MULTISINGLERAINBOW,
```
rogers-j: Please make separate commits for separate changes.

public/js/dev/game/init.js
```diff
@@ -345,7 +345,7 @@ function drawPlayers() {
             height: constants.HAND_POS[nump][j].h * globals.ui.h,
             rotation: constants.HAND_POS[nump][j].rot,
             align: 'center',
-            reverse: j === 0,
+            reverse: (lobby.reverseHand ? j != 0 : j === 0),
```
rogers-j: This isn't the same 'branch' as the other code. I suggest not modifying it at the same time as the other branch.

public/js/constants.js
jeep: Yeah, Zamiel asked me to just submit it despite it not being tested or reviewed. He said he would fix it up.

public/js/dev/game/init.js
rogers-j: There's another bit of code we need to reverse the gradient

PR #663, opened 2019-07-16
Add dump_game.py for dumping game records

nealmcb:
Still lots more features to add, and ideas to describe for improving game record format, but here's a start.

### DIFF
--- a/scripts/python/dump_game.py
+++ b/scripts/python/dump_game.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+"""
+dump_game
+~~~~~~~~~
+
+Parse, analyze hanabi.live games
+Game format documented at
+  https://github.com/Zamiell/hanabi-live/blob/master/misc/example_game_with_comments.json
+
+Retrieve games via e.g. wget https://hanabi.live/export/46008
+
+%InsertOptionParserUsage%
+
+Examples:
+  dump_game.py 46008
+  dump_game -h
+
+
+TODO:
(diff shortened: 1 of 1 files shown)


Zamiell:
this converts JSON to human-readable format? that doesn't strike me as particularly useful.
if you want a human readable output of a game, is it better to just extend the real API, e.g. `hanabi.live/export/123456/human`

nealmcb:
It would indeed also be useful to have the API generate other forms of human readable output.

But my ultimate goal is different. I see this PR as an early step in establishing an interoperable ecosystem for exploring Hanabi games among humans and AI agents, for integrating with hhttps://github.com/deepmind/hanabi-learning-environment et al.  I think we need to first firm up a robust, interoperable game record format, and that format needs utilities for dumping games, importing them into other systems, analyzing them, etc.

I know Python, not Go, so I'm doing what I can in Python.

I made it a PR here since so far it is specific to the hanabi-live format, and so at least people would be able to find it.

Of course, if you don't want to merge it, it could eventually live elsewhere.

Zamiell:
>I think we need to first firm up a robust, interoperable game record format
Do you see any current issues with the current game record format used in the Hanabi Live export API?

Zamiell:
> I know Python, not Go, so I'm doing what I can in Python.

Ok, but I mean even without knowing Go it would be pretty trivial to create the human API output, just copy paste the existing function and change how it is printed out.

PR #779, opened 2019-11-16
Added Omni-Ones and Null-5s variants

jakestiles:
I wouldn't be surprised if there was some logic shortcomings in the HanabiCard portion, but the rest was straightforward


public/js/src/game/ui/HanabiCard.js
```diff
@@ -397,7 +403,19 @@ export default class HanabiCard extends Konva.Group {
                         (suit) => suit.clueRanks === 'all',
                     );
                 }
-
+
```
Zamiell: this looks fucked

Zamiell:
remind me to implement omni everything at some point after i do some other work refactoring

PR #808, opened 2019-11-26
Fixed max scores global stats

### LINKED ISSUE #807: Max score global stats are incorrect
First, just wanna say thanks for maintaining this thing! My friends and I play using it each week and it's a fun way for us all to stay in touch. So thanks again for your efforts!

I found an issue as I was looking at [the global stats page](https://hanabi.live/stats) earlier today. I just couldn't make sense of the max scores:

Screen Shot 2019-11-25 at 10 06 16 PM

I was confused as to why "Total max scores achieved" was a fraction instead of an integer. And how is it possible all the player size variants have the exact same denominator? After digging into the code, I found [the] [problematic](https://github.com/Zamiell/hanabi-live/blob/39742b364b1ada9f3e185057e3484a0684d2ee7c/src/views/stats.tmpl#L44-L67) [files](https://github.com/Zamiell/hanabi-live/blob/39742b364b1ada9f3e185057e3484a0684d2ee7c/src/httpScores.go#L21-L24) and put together PR #808 for you to look at.

I apologize in advance for not following all of your setup instructions and getting things running locally. I've never used Go and honestly don't have the time right now to learn how to get it all set up on my machine and learn how to code properly in it. But hopefully the code is a useful starting point and someone can take it to the finish line.

Thanks again!

jwngr:
The code has four conflictingly named variables when keeping track of max scores. But ultimately what we need is two: one to track total max scores (`NumMaxScores`), and one to track and array of max scores for each person size variant (`NumMaxScoresPerType`).  Getting rid of the conflictingly named variables makes the template variable usage more clear.

Closes #807.

### DIFF
--- a/src/httpStats.go
+++ b/src/httpStats.go
@@ -18,9 +18,7 @@ type StatsData struct {
 	NumGamesSpeedrun      int
 	TimePlayedSpeedrun    string
 	NumMaxScores          int
-	TotalMaxScores        int
 	NumMaxScoresPerType   []int
-	TotalMaxScoresPerType int
 
 	Variants []VariantStats
 }
@@ -148,9 +146,7 @@ func httpStats(c *gin.Context) {
 		NumGamesSpeedrun:      globalStats.NumGamesSpeedrun,
 		TimePlayedSpeedrun:    timePlayedSpeedrun,
 		NumMaxScores:          numMaxScores,
-		TotalMaxScores:        len(variantsList) * 5, // For 2 to 6 players
 		NumMaxScoresPerType:   numMaxScoresPerType,
-		TotalMaxScoresPerType: len(variantsList),
 
 		Variants: variantStatsList,
 	}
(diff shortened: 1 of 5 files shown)


src/httpStats.go
```diff
@@ -18,9 +18,7 @@ type StatsData struct {
 	NumGamesSpeedrun      int
 	TimePlayedSpeedrun    string
 	NumMaxScores          int
-	TotalMaxScores        int
 	NumMaxScoresPerType   []int
-	TotalMaxScoresPerType int
```
chadheise: Alternatively, this could be  `numGamesPerType []int` I think it is more useful to see how many games were perfect out of all the games of that type than to see how many, of all perfect games, were with a certain number of players.

src/views/missingScores.tmpl
```diff
@@ -8,27 +8,27 @@
 <ul>
     <li>
         <span class="stat-description">Total max scores:</span>
-        {{.NumMaxScores}} / {{.TotalMaxScores}}
+        {{.NumMaxScores}}
```
chadheise: This could also be made out of the total number of games to add additional context.

PR #829, opened 2019-12-12
Add eslint dependencies to fix warnings and update generated go files

chadheise:
I saw warnings that transitive eslint packages needed to be manually added so explicitly added them. I also included the go files that were updated as a result of running ./install/install_dependencies.sh


Zamiell:
im not 100% sure that those packages are needed because hanabi live doesn't use JSX or React whatsoever

but ill just take your word for it and merge it for now i guess

PR #844, opened 2020-01-30
Add Null 5s variants (3, 4, 5, 6 suit)

jakestiles:
Before when I tried this, I tried adding omni 1s at the same time, which was based on the currently faulty prism 1s logic, so it was not merged. 

Now, it's just the null 5s, so I have more confidence it can be merged with few or no issues needing to be fixed before the variant is playable.

### DIFF
--- a/docs/VARIANTS.md
+++ b/docs/VARIANTS.md
@@ -33,6 +33,7 @@ Special:
 | Colors share suit | Dual color            |
 | All color clues   | Prism 1s (1s only)    |
 | All rank clues    | Multi-Fives (5s only) |
+| No clues          | Null-Fives (5s only)  |
 
 <br />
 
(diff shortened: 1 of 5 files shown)


jakestiles:
The checks failing don't have to do with the code I submitted I believe.

Zamiell:
i can finally consider merging this now that prism cluing is fixed

however i want to revamp the variants with consistent naming.
e.g. instead of calling it Multi 5's i want to call it Rainbow 5's.
same treatment goes to prism 1s.
and then i want to add every suit type to both cards.

jakestiles:
But this is null 5. Why do other unrelated variants need to be renamed before this is merged? Is there anything I can do?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Zamiell:
id rather implement them all at once instead of implementing *just* null 5s

Zamiell:
im working on this today, you can expect to see 400+ new variants soon

PR #849, opened 2020-02-21
Disable player time tooltip in speedrun games

### LINKED ISSUE #815: Feature request: remove "time taken" popup in speedruns
When you hover your mouse over anyone's name, you get a "time taken" popup.
I would like to remove this pop-up in speedruns.
The reason is that this pop-up appears on top of cards, and then left/right clicking the pop-up will not register as clicking the card.

See for example: https://www.speedrun.com/hanabi/run/m31k74wz around 0:15.

fpvandoorn:
I did not check whether this code works or compiles.

Closes #815.


fpvandoorn:
ping @Zamiell

Issue #852, opened 2020-03-02
Documentation for "efficiency" is inconsistent with its implementation


sardine8041:
In the UI tooltip and in `docs/FEATURES.md`, it says efficiency is calculated as `number of clues given / (number of cards played + number of unplayed cards with one or more clues "on" them)`, but in game this appears to be calculated as the reciprocal: `(number of cards played + number of unplayed cards with one or more clues "on" them) / number of clues given`.

For example, in a new game, on the first turn hint another player about 2 red cards in their hand. Clues given = 1, number of cards played = 0, number of unplayed cards with one or more clues "on" them = 2. According to the documentation efficiency should be 1 / (0 + 2) = 0.5, but the game will say it's 2.00.

Zamiell:
can you submit a pr

Zamiell:
thanks

PR #853, opened 2020-03-02
Fix documentation on "efficiency" in FEATURES.md and UI tooltip.

sardine8041:
See https://github.com/Zamiell/hanabi-live/issues/852.


Zamiell:
thank you sir!

PR #855, opened 2020-03-08
Stop the stack bases from flying around during replays

kemurphy:


Zamiell:
this bug is already fixed on master, did you test it?

rogers-j:
For what it's worth, I can reliably repro on the live site by middle-clicking a card after it has been clued.

Zamiell:
the live site is on a commit from like 6-12 months ago

kemurphy:
Indeed, this isn't a bug on master.  Why don't we keep the live site up to date?

Zamiell:
right now the site is on an older commit temporarily because there are some problems remaining on the master branch that i still need to fix

kemurphy:
Would these problems be the typecheck errors on master?

Zamiell:
ya

Zamiell:
yeah

Issue #908, opened 2020-04-02
Access all game histories


aalok-sathe:
Is there a way to access all game histories of all games played online (or however many that are stored in your database)? I'm a board games enthusiast and this would be for small-time research as a college project towards building a baseline Hanabi AI. Could be integrated into this project someday.

Zamiell:
easily? no

but database ids are sequential. so you can just send a replayCreate command for IDs 1 through N

Zamiell:
and make sure to throttle it quite a bit so that you dont kill my web server = )

aalok-sathe:
of course! i will take care of that. how do i send that command?

Zamiell:
you'll have to write a script to emulate a browser client, login to the websocket server, etc

Zamiell:
closing this as access to the database won't be officially supported

PR #909, opened 2020-04-03
Increase MaxAge of session cookie

rmelick:
Increase MaxAge of session cookie to make it easier for non-javascript clients to create the websocket connection.  Update references that mention the cookie duration.

### DIFF
--- a/src/http.go
+++ b/src/http.go
@@ -89,7 +89,7 @@ func httpInit() {
 		// After getting a cookie via "/login", the client will immediately
 		// establish a WebSocket connection via "/ws", so the cookie only needs
 		// to exist for that time frame
-		MaxAge: 1, // in seconds
+		MaxAge: 15, // in seconds
 		// Only send the cookie over HTTPS:
 		// https://www.owasp.org/index.php/Testing_for_cookies_attributes_(OTG-SESS-002)
 		Secure: true,
--- a/src/httpLogin.go
+++ b/src/httpLogin.go
@@ -1,7 +1,7 @@
 /*
 	This is only part 1 of 2 for login authentication.
 	The user must POST to "/login" with the values of "username" and "password".
-	If successful, then they will receive a cookie from the server with an expiry of 5 seconds.
+	If successful, then they will receive a cookie from the server with an expiry of 15 seconds.
 	The client will then attempt to make a WebSocket connection; JavaScript will automatiaclly
 	use any current cookies for the website when establishing a WebSocket connection.
 	The logic for opening a websocket connection is contained in the "websocketConnect.go" file.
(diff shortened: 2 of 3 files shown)


Zamiell:
why exactly would a robot need a longer time to create a connection than a human would?

rmelick:
It seems to be taking longer than 1 second for my java code to get the response from the /login endpoint and run through its own code before making the call that opens the websocket.  With general network latency, I would expect at least 200-500ms for my code to get any response from the login endpoint, which only gives it 500ms to make another call.  

The http client I'm using actually checks that the cookie is still valid before making the request to the server.  My guess is that the javascript hanabi live client doesn't actually enforce that the cookies are unexpired so it doesn't run into this issue.

Zamiell:
what is your ping to hanabi.live? do you live in antarctica or something?

rmelick:
Checked my ping, it's only about 50ms.  

So, I looked more into why my java client was treating the cookie as already expired before trying to make the request.  When it parses the cookie, it looks at the `Expires` field and compares to the current time see how many seconds the cookie will remain valid.  It takes the millisecond difference and divides by 1000 to get the difference in seconds.
```
(cal.getTimeInMillis() - whenCreated) / 1000;
```
https://github.com/openjdk/jdk/blob/17ff85d4ada184b65f091251820d5b0e0b3e4164/src/java.base/share/classes/java/net/HttpCookie.java#L1067

Since the backend is setting the cookie to expire exactly 1 second after it was created, there is less than 1000 milliseconds of validity remaining by the time I do this check, which means the integer division gives a result of 0 seconds remaining.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.17.2

So, I think even having a server side expiration of 2 seconds would probably meet my need, but it felt like 15 seconds would also make it easier for others who wanted to build their own clients.  They would be able to use slower internet connections, or slower client code, etc.

Zamiell:
> Since the backend is setting the cookie to expire exactly 1 second after it was created, there is less than 1000 milliseconds of validity remaining by the time I do this check, which means the integer division gives a result of 0 seconds remaining.

interesting. sounds like Java sucks LOL
i'll change it to 2 seconds, which will become active the next time the server restarts.

by the way, what is your bot doing? is it related to #908? Java does not seem like a very good choice of a programming language for a Hanabi bot. why did you choose Java over e.g. Node? Node seems like the best choice for this, as you could just copy all of the relevant code from the real client. Kotlin, Golang, and Python all seem like good choices as well.

> Is there a reason that 1 second expiration is helpful for the normal use case?

the cookie is supposed to be ephemeral, thus i engineered it to have the lowest possible value for security purposes

Zamiell:
> Is there a reason that 1 second expiration is helpful for the normal use case?

another reason is that if the cookie is N seconds long, then the user cannot refresh the page any more than 1 time every N seconds, so minimizing N is desired

rmelick:
I picked java to make it easy to integrate with the bots people wrote for the IEEE competition: (http://hanabi.aiclash.com/domain.html, https://ieeexplore.ieee.org/document/8848008)

PR #940, opened 2020-04-13
use Prettier for formatting

jtbandes:
[View diff ignoring indentation changes (easier to read)](https://github.com/Zamiell/hanabi-live/pull/940/files?w=1)

- Install prettier and the prettier eslint plugin (so that `lint.sh` will include it)
- Add .vscode/extensions.json with recommended extensions

Disadvantage: all the changes in this PR. I admit Prettier makes some less than perfect choices especially around binary operators.

Advantage: turn on Format On Save in VSCode, never have to format manually again


jtbandes:
> until it can actually conform fully to the airbnb guide

what's the motivation for conforming so strictly to the guide? The eslint configs/plugins provided by prettier ensure that there are no conflicts with the rules, so the prettier output will pass lint.

Zamiell:
well, airbnb is the most popular javascript style guide in the world, so i want the hanabi live code to look like all the other javascript code in the world. similar to how when i write go, i want it to look like all the other go code in the world, and when i write python, i follow pep 8. what's the point of a style guide if you don't follow it?

of course, to be fair, I have disabled some specific airbnb rules already. but those are few and far between, and each disabled rule has been carefully considered, typically making the actual quality of the code worse, as opposed to just the style.

Zamiell:
maybe a middle ground is to configure vscode to use `eslint --fix` on save? I've not looked into that. obviously it wouldn't be as robust as prettier, but it would probably do the most important things, off the top of my head being auto-indent and auto-adding-semicolons.

Zamiell:
as an aside, im noticing yet another problem with the prettier results, they are constructing if statements that break the line length rule by a few characters. maybe the code isn't counting the trailing " {" as part of the line or something?

jtbandes:
It looks like there actually is a setting for [`trailingCommas: "all"`](https://prettier.io/docs/en/options.html#trailing-commas) which I will enable when I get a chance.

> they are constructing if statements that break the line length rule by a few characters 
You mean the `printWidth: 120` I set (which was an arbitrary choice) or some other length rule?

I do also like eslint --fix on save, but eslint doesn't do nearly as much basic whitespace/formatting as prettier does. Well, unless you configure the prettier plugin, then your eslint --fix will (I think) rerun prettier as well.

There is some discussion about airbnb style vs prettier at https://github.com/airbnb/javascript/issues/1548

And here's some popularity data https://www.npmtrends.com/eslint-config-airbnb-vs-prettier-vs-standard-vs-eslint-config-prettier-vs-eslint-plugin-prettier

Zamiell:
> You mean the printWidth: 120 I set (which was an arbitrary choice) or some other length rule?

airbnb prescribes max line length of 100 chars (19.13)

> There is some discussion about airbnb style vs prettier

that was a pretty interesting read. it sounds like the tooling for airbnb+prettier still needs a lot of work. but that thread was 2 years ago, maybe it has progressed since then?

> And here's some popularity data

that's also interesting but i don't think it necessarily shows that people are using prettier and airbnb together. people could theoretically use prettier with any arbitrary style guide, so i would expect the numbers for prettier to be above any single style guide. and the eslint-config/plugin just shows that people are using eslint alongside prettier (which again isn't particularly relevant because you can use eslint with any arbitrary style guide), correct me if i am wrong.

what I think it does show is that more people adhere to airbnb style over standard style, which was one of my main points

Zamiell:
looks like people are pissed about prettier's objectively bad handling of suffix conditions, with 2 years of not fixing the issue, causing people to use forks:

https://github.com/prettier/prettier/issues/3806

Zamiell:
i added eslint --fix on save: https://github.com/Zamiell/hanabi-live/commit/58040af16883bb371fc2a2d5bf883f3ff148e203

Zamiell:
i've decided to switch the project to prettier now for various reasons

PR #961, opened 2020-04-18
Add a docker configuration for production

tkw1536:
The existing Dockerfile is made for a development environment. In particular to run the docker-compose.yml file it is required to manually compile the frontend dependencies on the host machine. Furthermore it runs the hanabi-live daemon as root inside the docker container, which
is a security issue.

This PR adds a Dockerfile and docker-compose.yml for a production setup in the 'docker' folder. This does not have any dependencies on the host environment and can be built on a server having only a checkout of this repository and the docker-compose executable. The Dockerfile uses a common docker practice of "multi-stage builds" to achieve this. It also mitigates most security risks by running hanabi-live as a limited user within Docker.

A new section on how to start this docker image has been added to the installation documentation.

Furthermore, this PR also adds a "smoketest" for this new setup to .travis.yml. This smoketest sets up a clean environment, builds the docker image, and then starts it. Afterwards it checks if the root URL of the server returns HTTP 200.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,5 +6,18 @@
 *
 
 # Allow files and folders with a pattern starting with !
-!.env_template
+
+
+# for building the docker-entrypoint
+!docker/src
+
+# for building the backend
 !src
+
+# for building the frontend
+!.git
+!public
+!build_client.sh
+
+# for the development docker container
+!.env_template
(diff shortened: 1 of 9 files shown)


Zamiell:
lgtm

what is the motivation behind this PR? are you running your own version of hanabi live?

tkw1536:
Yes, something like that. The original motivation was hanabi.live going down for maintenance while I wanted to play with a few people from work. So I setup a version of this in about 30 minutes. I now spent some time cleaning this up, and making it easier to deploy for the future.

Zamiell:
ok.

btw i modified the functionality of the site so that you can still create games while it is in the rebooting phase, which lessens the impact

PR #1036, opened 2020-05-12
Fix empathy in special suits

padiwik:
Should fix current outstanding empathy issues. Negative color clues now
give negative rainbow even in special-ones/fives. Rank clues can now
result in crossed-out suit pips even in pink variants.


Zamiell:
this breaks null-ones variants apparently

Issue #974, opened 2020-04-21
https://boardgamegeek.com/boardgame/290357/hanabi-grands-feux


Zamiell:

robertos:
I'd be interested in taking a stab at implementing this. @Zamiell I imagine you're talking about the Black Powder variant?

If so, I think it could be developed in parts:
1. Add the ability to have a suit be reversed (three 5's and one 1, must be played 5 -> 1). That would allow already for hundreds of new variants since Reversed (or "Upside-Down") is a property like Dark or Null.
2. In this particular case, the variant is "Reversed White (6 suits)". Optionally, Reversed White can be renamed or aliased in the list as "Black Powder" as a special case. 
3. There's a special scoring, which is to make the score start at -5 (i.e. max score is 25 even though there are 6 suits, each card not played from the special suit is a -1). I don't think this is necessary, since it's just changing the initial score and doesn't affect the game at all.
4. If desired, reversed suit(s) can have some visual treatment / icon. 

Rules (in French): https://www.philibertnet.com/en/index.php?controller=attachment&id_attachment=3851

Zamiell:
sure, go for it

Zamiell:
its faster to talk to me in discord btw

robertos:
OK, I'll find you there if/when I get to this (BTW, this was definitely fast enough 😄)

Zamiell:
right now i think master is bugged, you might want to fork from the version from 4 days ago

Zamiell:
done-ish

PR #1109, opened 2020-05-26
Add reversed suits

robertos:
This PR adds the mechanisms for Reversed suits, and adds the simplest Reversed variants (3-6 suits).
This is the first step to enable the Black Powder official variant (i.e. White Reversed) and other derivatives.
Also refactor the "Up and Down" variant code to reduce comparing with copies of the variant name, and reuse code between "Up and Down" and any "Reversed" variants.

**TODO:** better icons/visual treatment when playing with Reversed suits.


robertos:
Tagging issue #974

Zamiell:
i think it does make sense to borrow the color teal for the reversed suit. however, the reversed suit should not have a hashtag pip. this is confusing because it overlaps with pink, implying that the reversed suit has the same properties as pink, which it doesn't.

we'll need to come up with a new pip. i propose something like this: https://fontawesome.com/icons/history?style=solid

can you add that?

Zamiell:
in other news, the problem gets a little bit more complicated when we consider how this should work with a hypothetical Reversed Rainbow suit.

a simple way to resolve this problem is to have the Reversed Rainbow suit have the appearance of a normal rainbow suit, but with a "reverse" pip to signify that it is the reversed version.

Zamiell:
future thoughts:

ideally, we would want a reversed version of every suit.
and that would mean duplicating every "special" suit in the suits.json, which would be silly.

perhaps better to create the suits dynamically in suitsInit.js, meaning that for every suit that has `createVariants: true`, then add a reversed version of that suit to the suits map (e.g. so that it is done in code rather than blowing up the size of the suits.json file).

robertos:
Thanks for the comments!

> i think it does make sense to borrow the color teal for the reversed suit. however, the reversed suit should not have a hashtag pip

For sure, the color can be anything, but the icon was a placeholder. I'm not a huge fan of the different pip because of the different variants of the suit, as you said (e.g. Dark Rainbow Reversed, or even 2 reversed suits). For example a potentially interesting variant would be 3 reversed and 3 regular suits, but then half of the cards would have the same pip.

Do you think it's important to allude to the original name of the variant, Black Powder? Does it make sense for the pip to be related to powder? Otherwise the clock works, I also like a down arrow like [this](https://fontawesome.com/icons/sort-amount-down?style=solid)

> we would want a reversed version of every suit

I like that idea, we can generate the suits as you proposed.

What do you think is the minimal set of features/polish to get this PR in and ready for people to try? I'll throw a few, in what I think is a priority order:
- [ ] Choose a new pip for the Reversed suit and finalize the color
- [ ] Figure out a better presentation for the stack direction than "Up" and "Down" labels. I think nothing and "Reversed" could work well.
- [ ] Manually add a few new suits for basic Reversed variants (Black, White, Rainbow, Dark Rainbow, Gray, Pink)
- [ ] A system to auto-gen reversed versions of any suit and generate the variants appropriately
- [ ] A system to allow users to choose any number of suits to be reversed in any variant (except "Up or Down") during game creation

Zamiell:
> Do you think it's important to allude to the original name of the variant, Black Powder?

not really tbh.
we have already departed from the "cannon" by using different colors.
(our colors make more sense the official versions because e.g. white is the absence of all colors so a colorless suit should be white)
just document that whatever name we chose is the same as "Black Powder" in features.md and leave it at that

> Does it make sense for the pip to be related to powder?

probably not

> Otherwise the clock works, I also like a down arrow like this

i have no strong preference, whatever you think will look best

> Figure out a better presentation for the stack direction than "Up" and "Down" labels. I think nothing and "Reversed" could work well.

i have an idea for this, i could make an "up" and "down" symbol similar to the kt thing.
in reversed, it would *only* show for the reversed suit though. (in up or down, it would show for everything)

> Manually add a few new suits for basic Reversed variants (Black, White, Rainbow, Dark Rainbow, Gray, Pink)

not yet

> A system to auto-gen reversed versions of any suit and generate the variants appropriately

yeah just some add code to the python script if you don't mind, since i will be using that script eventually

> A system to allow users to choose any number of suits to be reversed in any variant (except "Up or Down") during game creation

a "create your own variant" system is outside of the scope of this pr, this is a super complicated feature that has long been requested and hasnt been implemented for various reasons

Zamiell:
>  I'm not a huge fan of the different pip because of the different variants of the suit, as you said (e.g. Dark Rainbow Reversed, or even 2 reversed suits). For example a potentially interesting variant would be 3 reversed and 3 regular suits, but then half of the cards would have the same pip.

propose a system then that solves this in an adequate way

robertos:
Sounds good. 

> propose a system then that solves this in an adequate way

How about adding a down arrow next to each pip and downscaling the pips a bit? I can try that.

Will give a stab at that then. So, for this PR, the min-bar is:
- [ ] Choose a new pip for the Reversed suit and finalize the color (I'll try both and see what works)

Stretch goals:
- [ ] System to generate suits for reversed suits (on .py script)
- [ ] System to auto-gen pips for reversed suits

Out of scope here, for the future
- [ ] Actually generating other reversed suits/variants
- [ ] "Up" and "down" symbol under stacks instead of text labels
- [ ] Fully custom variants

Zamiell:
> How about adding a down arrow next to each pip and downscaling the pips a bit? I can try that.

yeah i like that
depending on if it looks good i guess =p

Zamiell:
and thanks for the pr so far!

Zamiell:
btw, check out the images in the pips folder, and read this file:
https://github.com/Zamiell/hanabi-live/blob/master/public/img/pips/notes.txt

robertos:
Option 1 (arrow):

image

robertos:
Option 2 (clocks):

image

Zamiell:
i kinda like option 2

and maybe the "modifier" for a suit pip can be the circle surrounding it as a border

rogers-j:
Maybe I assign too much importance to the fact that the backward clock icon is already part of the UI.

rogers-j:
They both seem a bit noisy at low resolutions. For the former, maybe just three horizontal bars?

rogers-j:
Adding a downward-pointing arrow on the left side of any suit image seems distinctive enough.

Zamiell:
here's an image of what i was talking about earlier.
for example, in up or down:

image

both arrows would be set to invisible by default.
and then, when a stack is activated as going up, it enables the up arrow, and vice versa if going down

Zamiell:
maybe white is better:

image

rogers-j:
It doesn't need to take all the space:

image

rogers-j:
Or maybe you'd prefer

image

or 

image

Zamiell:
on the contrary, there is an argument for making it obnoxiously big because it is super-easy to make the mistake of misremembering the stack direction while playing an up or down game

rogers-j:
Also keep in mind that the arrows are going to be a bit arbitrary in Keldon mode when cards rotate.

Zamiell:
i had intended for the arrows to only show on the stacks. but i guess they could also show on all cards too

rogers-j:
When the stacks are shown isn't the direction still undefined in the existing up-or-down variant?

rogers-j:
Perhaps graphically (but not programmatically) the convention should be that 'up' is treated as a default until 'down' is established.

Zamiell:
as i said above, the arrows would only appear when the stack direction is already decided

rogers-j:
Right, and in up-or-down the direction is the direction not decided by playing something onto the empty stack?

Zamiell:
if the stack is empty, then the direction isn't decided yet

rogers-j:
Okay, it seems like we're agreeing.

rogers-j:
Anyway, back to icons. Maybe something more along these lines?

image

It scales down a bit better

image

Zamiell:
ok i programmed it for real, looks like this:

image

robertos:
Looks pretty good! What visual do you wanna go with? I also was going to try using a simple arrow as the pip https://fontawesome.com/icons/arrow-alt-circle-down?style=solid

Zamiell:
yeah that could look better at small resolutions

robertos:
Option 3 - arrow

image

robertos:
Option 4 - chevrons

image

robertos:
Here's another few mockups for the generic arrows that won't intersect with any pip 1-5

image image 
image image

robertos:
On the mockup above, I think the arrows on the bottom left work best since the "notes" icon is on the top right

Zamiell:
very interesting, i like both option 3 and option 4

robertos:
My favorite is the [generic arrow on the bottom left](https://user-images.githubusercontent.com/621574/83052805-acf0d880-a004-11ea-8118-7194a72b9843.png), since it can be applied to any suit. 


An alternative that was suggested by my group when I showed the options was to have the arrow replace the bottom right number, since that number is not really necessary for the game. Something like this

image

And on up-or-down you'd also need the up version
image

robertos:
If we're using generic arrows, one thing to consider is that the UI already has an arrow (the current turn pointer), so we should probably use the same for consistency

image

Zamiell:
how about this

image

Zamiell:
ill try with removing the 2nd number

robertos:
> how about this

I really like it, but I think the arrows should be a non-suit color (e.g. white) to be clearer. I also think the arrows should always be on the same side (right or left) regardless of up or down. My favorite locations are bottom right (replacing number) or bottom left.

Zamiell:
image

image

i kinda like it better with the colors

Aigul9:
colored arrows look more appropriate to the background and do not distract. can't we just pay more attention to the game without making everything oversimplified?

Zamiell:
relax

Aigul9:
)) let's make a bot which will highlight finessed or bluffed cards (even in the layer if cards don't match) automatically, according to hyphen-ated conventions, show us cards with/without priority and suggest the best clue from players' pov

Zamiell:
ok this is live now in https://github.com/Zamiell/hanabi-live/commit/08e93c6a6152eacf0ae21da9b0e8eb410341b852

robertos:
@Zamiell Awesome. I'll pull that into this branch tonight and test it out. In that case the new Reversed suit will simply use the Teal color and pip (in prep for auto-generating Reversed suits)

Zamiell:
sure

rogers-j:
One alternative for Zamiel to consider is only drawing the arrows _on the played pile in the center_, but not the cards in hands (or the discard, which will be especially crowded!)

They wouldn't even need to be part of the cards themselves, just something that draws on top of the pile.

Zamiell:
i like it better on all the cards

robertos:
Finally got to it! Commit coming soon. It auto generates all reversed suits :)

image

Zamiell:
wewt

public/js/src/game/ui/variants/reversible.ts
```diff
@@ -17,6 +17,17 @@ export const hasReversedSuits = () => {
...
 
+export const shouldShowArrow = (direction: number) => {
+  if (!hasReversedSuits()) {
+    return false;
+  }
+
+  // The arrow should be shown when the suit is reversed, or
+  // whenever it's determined in "Up or Down"
+  return (direction === STACK_DIRECTION.DOWN)
+      || (direction === STACK_DIRECTION.UP && isUpOrDown());
```
Zamiell, line 28:
```
return (
  direction === STACK_DIRECTION.DOWN)
  || (direction === STACK_DIRECTION.UP && isUpOrDown()
);
```

src/command_table_start.go
```diff
@@ -248,6 +248,9 @@ func commandTableStart(s *Session, d *CommandData) {
 		}
 	}
 
+	// Record the initial status of the game
+	t.NotifyStatus()
```
Zamiell: why is this getting moved

robertos: It's been un-moved on the latest commit. I did separate out the NotifyStackDirections since that is needed before the cards are drawn

robertos: Otherwise any cards on player's initial hands will have white arrows.

public/js/src/game/ui/variants/reversible.ts
robertos: Sure

PR #1132, opened 2020-05-29
Fix condition that differentiates up or down from reversed for max score

### LINKED ISSUE #1129: Maximum score issue in up or down variant
Haven't had time to do any extensive testing, but the up or down variant seems to be miscalculating the maximum score.

My understanding is that the game below should have continued and the maximum score is still 30. 

Screenshot from 2020-05-28 16-47-50

I should note that the "womp womp" sound played every time we discarded 1s or 5s, independent of whether those cards were required to finish the game.

robertos:
Fixes #1129


Zamiell:
woot

PR #1154, opened 2020-06-01
Make actions and constants strongly typed. No functional changes

### LINKED ISSUE #1153: convert stuff in constants.ts to enums
including currentScreen to an int or whatever

robertos:
Fixes #1153 

Also changed some interfaces to straight up types. Didn't do this replacement everywhere though.
Main benefit: now actions are strongly typed, so it won't compile with typos on action names.

### DIFF
--- a/public/js/src/constants.ts
+++ b/public/js/src/constants.ts
@@ -27,39 +27,40 @@ export const ARROW_COLOR = {
 export const FADE_TIME = 350; // In milliseconds
 
 // These constants much match their server-side counterparts
-export const ACTION = {
-  PLAY: 0,
-  DISCARD: 1,
-  COLOR_CLUE: 2,
-  RANK_CLUE: 3,
-};
-export const CLUE_TYPE = {
-  COLOR: 0,
-  RANK: 1,
-};
-export const REPLAY_ACTION_TYPE = {
-  TURN: 0,
-  ARROW: 1,
-  SOUND: 2,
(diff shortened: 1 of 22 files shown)


public/js/src/game/ui/action.ts
```diff
@@ -58,7 +60,7 @@ export default (data: any) => {
 };
 
 // Define a command handler map
-const actionFunctions = new Map();
+const actionFunctions = new Map<ActionIncludingHypothetical['type'], any>();
```
robertos, line 63: #magic

public/js/src/constants.ts
```diff
@@ -27,39 +27,40 @@ export const ARROW_COLOR = {
...
-  EFFICIENCY: -4,
-  MIN_EFFICIENCY: -5,
-};
-export const STACK_DIRECTION = { // Used in the "Up or Down" and "Reversed" variants
-  UNDECIDED: 0,
-  UP: 1,
-  DOWN: 2,
-  FINISHED: 3,
-};
+export enum ActionType {
```
Zamiell, line 30: airbnb guide says that constants should be uppercase with underscores

robertos: This is not a constant anymore. It's a type now

robertos: https://basarat.gitbook.io/typescript/styleguide#enum

Issue #800, opened 2019-11-22
remove the clues part of the status message


Zamiell:

Zamiell:
@robertos can i close this now?

robertos:
We haven't been using the clues for a little while. Maybe we need to test more edge cases like detrimental chars before nixing this

Zamiell:
done

robertos:
Nope, not done yet... https://github.com/Zamiell/hanabi-live/blob/card-reducer/server/src/table_notify.go#L142

Zamiell:
closing this in favor of #710

PR #1168, opened 2020-06-03
Remove clues from status message

robertos:


robertos:
Hypo feels hacky, since it doesn't know about state and reimplements game logic

robertos:
Fixes #800

Zamiell:
yes hypo is a short-term solution.
in the long-term, the client will do all the game logic, and then the hypo will just use that.

client/src/game/ui/action.ts
```diff
@@ -440,42 +439,18 @@ actionFunctions.set('stackDirections', (data: ActionStackDirections) => {
...
-      // Show a yellow border around the discard pile
-      // (to reinforce that this is a "Double Discard" situation)
-      globals.elements.noDiscardBorder!.hide();
-      globals.elements.noDoubleDiscardBorder!.show();
-    } else {
-      globals.elements.noDiscardBorder!.hide();
-      globals.elements.noDoubleDiscardBorder!.hide();
-    }
+  // Update double discards
+  if (globals.clues !== MAX_CLUE_NUM
```
Zamiell, line 446: if (
  globals.clues

robertos: I believe this kind of rule can be enforced with eslint

Zamiell: which rule?

robertos: I'm checking, not sure

public/js/src/game/ui/sideEffects.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/prefer-default-export */
```
Zamiell: this shouldn't be needed bcuz you have 2 exports, right?

client/src/game/ui/action.ts
robertos: Hm couln't find it.

public/js/src/game/ui/sideEffects.ts
robertos: Nice catch - leftover from development. Will remove

PR #1174, opened 2020-06-03
feat: Add AllOrNothing logic to game code

ikelos:
This patch:
* uncomments the AllOrNothing option from the UI
* adds the logic to prevent gameplay after a critical discard or when the next player has no cards and no clues
* disables setting the endTurn variable to prevent the endTurn condition firing
* prevents the "last turn" text from being applied in allornothing games, when there are no more cards in the deck


Zamiell:
"This branch has conflicts that must be resolved"

can you rebase it on master? i changed the location of where the files live, so thats probably whats causing it

PR #1187, opened 2020-06-04
Make state changes pure using Lodash

robertos:
Also calculate score as it goes (score shouldn't be necessary on status)
No functional changes.


Zamiell:
thank you!

PR #1188, opened 2020-06-04
/random returns an integer

duckmammal:
proposing to make it clear that you'll get an integer if you use random

### DIFF
--- a/docs/CHAT_COMMANDS.md
+++ b/docs/CHAT_COMMANDS.md
@@ -13,7 +13,7 @@ If you need general help with the website, then read the [features page](FEATURE
 | /rules                   | Get the link for the [Hanabi Community Guidelines](https://github.com/Zamiell/hanabi-live/blob/master/docs/COMMUNITY_GUIDELINES.md)
 | /new                     | Displays a stock message for new users, encouraging them to join the Hyphen-ated group
 | /replay [game ID] [turn] | Generate a link to a replay so that you can share it with others
-| /random [min] [max]      | Get a random number
+| /random [min] [max]      | Get a random integer
 | /uptime                  | Get how long the server has been online
 | /timeleft                | Get how much time is left before the server shuts down
 | /shrug                   | ¯\\_(ツ)_/¯


Zamiell:
"integer" could be confusing for people who don't know what integers are / ESL

aliblong:
https://discordapp.com/channels/140016142600241152/333680903748321282/469911670400221215

Zamiell:
https://english.stackexchange.com/questions/29927/is-integer-a-commonly-understood-term/29930

aliblong:
A valid response might then be "let's call it 'whole number' instead of 'integer'" (if that's even the actual behaviour); hiding a reversion, then later doing the explicit reversion when you got called out, all without posting *any* explanation, was probably not the best approach.

Also, do you really think saving a small fraction of our playerbase from having to do a Google search (or double click then right click then left click) to get the definition of a word outweighs the uncertainty that has led to *multiple* instances of this exact question?

Zamiell:
as i seem to recall, we disagreed about whether "integer" could be confusing, but instead of discussing it further, you explicitly reverted my commit (fd66f21) with a rude commit message, so i didn't feel so bad about reverting it right back (8ebcd72), which seems like a kind of silly thing to do in retrospect (from both sides).

if you want to have an actual discussion about whether integer is confusing, that's fine, but lets not get into a pissing match where we revert each others commits.

as it stands, i feel like the stackexchange thread that i posted has a pretty good summary of the situation, and it looks like the vast majority of people who viewed this thread agree with this perspective.

aliblong:
> as i seem to recall, we disagreed about whether "integer" could be confusing, but instead of discussing it further, you explicitly reverted my commit (fd66f21) with a rude commit message

This is a misleading portrayal of events. As I already stated, you hid a reversion to my initial commit (a76712f) in a totally unrelated commit (7935b13) -- one that otherwise comprised entirely whitespace changes no less, making it that much less obvious. I suspect this is because you didn't want me to realize what was happening; the most generous interpretation is that it was just a severe and harmful lack of git hygiene. So yeah, my *follow-up* commit title was a little snarky for that reason... and because you already had a history of flaming me in commit messages and GitHub issues ("it's just a joke", as you liked to explain). Only at that point did you do an explicit reversion. All the while, you never made a statement in Discord or GitHub on why you took this course of action, including totally ignoring the poll I took regarding the issue, presumably because you had no legs to stand on and just wanted to squash the discussion

Now, back to this iteration of things, you still haven't addressed any of this:
> Also, do you really think saving a small fraction of our playerbase from having to do a Google search (or double click then right click then left click) to get the definition of a word outweighs the uncertainty that has led to multiple instances of this exact question? Not to mention, if someone has the inclination to inquire about the nature of the /random function, they already have a very strong context on what "integer" might be.

To preempt your argument, no, pointing to a SE thread where someone's mom doesn't know what an integer is when given 0 context doesn't count as addressing this. And I disagree with your comment here:

> it looks like the vast majority of people who viewed this thread agree with this perspective.

I assume you mean the SE thread. There, the second and third top answers state that they think "integer" is a commonly known term, and the comments across all answers are either neutral or agreeing that "integer" is a commonly known term. Hard to call that a vast majority in your favour. But it's mostly irrelevant, because the SE thread itself is only of low relevance.

Here are the facts:
* Multiple people have *explicitly* requested clarity on the `/random` function or signaled agreement with such a change: myself, MasN, and now duckmammal and florrat. Who knows how many more have silently wondered about the behaviour, or even taken it upon themselves to determine the behaviour through trial and error.
* Anyone viewing the documentation for `/random` has a strong context by which to infer the meaning of the words there.
* Anyone who goes to the trouble of looking up command documentation can perform a Google search as easily as breathing.

There is *zero* chance that someone will be confused or hurt as a result of accepting PR. My only quibble is that it should be even *more* precise, stating `"Get a random integer from the range [min, max)"` (again, assuming that's indeed the behaviour).

Zamiell:
historically, the /random feature has been used to get a random variant, e.g. "/random 1 6", or to pick a random starting player, e.g. "/random 1 5". lately, no-one has used the random feature at all, for any reason. so my intuition is that it would be silly to me to explicitly clarify in the documentation that doing a "/random 0.3 0.5" would be illegal. call me crazy, but i don't think im coming out of left field here.

> presumably because you had no legs to stand on and just wanted to squash the discussion

as I seem to recall, the "poll" was you just asking people to post emojis or something. after noticing the emoji poll in Discord, I went back to coding on the website, promptly forgot it 5 seconds later, and I haven't thought about the issue again for two years. there were no follow-up discussion on the poll from you or anyone else. i don't think portraying that as me "squashing consensus" is very charitable.

in actuality, I would say that changing the word "number" to "integer" is pretty insignificant and hardly worth my time to even evaluate  the pros and the cons. from the length of your reply, you seem to care deeply about this issue and deem it quite important, so I will trust your judgement and accept this PR as a measure of goodwill.

PR #1211, opened 2020-06-08
Fix #1210 - Revert to evt.which

### LINKED ISSUE #1210: shred replay leader can't right click score/eff/etc

robertos:
For now. Later, should change all which to button or buttons.


robertos:
Fix #1210

PR #1215, opened 2020-06-09
Refactor state reducer to be purely functional

robertos:
Move some types and functions around to disentangle them from the UI
variantUtils eventually will be replaced by the strategy pattern.
Destroy some any's for good measure.


client/src/game/ui/websocket.ts
```diff
@@ -369,18 +370,15 @@ commands.set('gameAction', (data: GameActionData) => {
...
+  // Update the game state
+  globals.state = stateReducer(globals.state, actionMessage);
 
   // TEMP: We need to save this game state change for the purposes of the in-game replay
   globals.replayLog.push(actionMessage);
 
   if (actionMessage.type === 'turn') {
     // Make a copy of the current state and store it in the state table
-    globals.states[actionMessage.num] = _.cloneDeep(globals.state);
+    globals.states[actionMessage.num] = globals.state;
```
Zamiell, line 381: does that make a copy?

robertos: Copy is not needed, can remove comment

PR #1232, opened 2020-06-13
Fix hypoRevealed

padiwik:
Now the toggle behaves properly

### DIFF
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -888,9 +888,10 @@ export default class HanabiCard extends Konva.Group {
   }
 
   // This card was either played or discarded (or revealed at the end of the game)
-  reveal(msgSuit: number, rank: number) {
+  reveal(msgSuit: number, msgRank: number) {
     // Local variables
-    const suit = msgSuitToSuit(msgSuit, globals.variant);
+    const suit = msgSuit === -1 ? null : msgSuitToSuit(msgSuit, globals.variant);
+    const rank = msgRank === -1 ? null : msgRank;
 
     // Set the true suit/rank on the card
     this.state.suit = suit;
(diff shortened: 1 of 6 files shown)


client/src/game/ui/HanabiCard.ts
```diff
@@ -909,9 +909,10 @@ export default class HanabiCard extends Konva.Group {
   }
 
   // This card was either played or discarded (or revealed at the end of the game)
-  reveal(msgSuit: number, rank: number) {
+  reveal(msgSuit: number, msgRank: number) {
     // Local variables
-    const suit = msgSuitToSuit(msgSuit, globals.variant);
+    const suit = msgSuit === -1 ? null : msgSuitToSuit(msgSuit, globals.variant);
```
robertos, line 914: I'd fix this in msgSuitToSuit since the same issue happens elsewhere (and remove the check there)
```
export const msgSuitToSuit = (
  msgSuit: number,
  variant: Variant,
) => {
  if (msgSuit < 0 || msgSuit >= variant.suits.length) {
    return null;
  }
  return variant.suits[msgSuit];
}
```

robertos: The opposite is also true, we're depending on the behavior of indexOf (returning -1 when not found) on suitToMsgSuit, might as well make it explicit.

padiwik: Noted, I was just copying some code from action.ts so I'll fix it there too

client/src/game/ui/HanabiCard.ts
```diff
@@ -939,6 +940,35 @@ export default class HanabiCard extends Konva.Group {
...
+  replayRedraw() {
+    if (globals.deckOrder.length === 0) {
+      return;
+    }
+    const suitNum = globals.deckOrder[this.order].suit;
+    const trueSuit = msgSuitToSuit(suitNum, globals.variant);
+    const trueRank = globals.deckOrder[this.order].rank;
+
+    if (
+      globals.hypothetical && !globals.hypoRevealed
```
robertos, line 954: inb4 Zamiel commetns that you should have one condition per line

padiwik: now they are...

PR #1247, opened 2020-06-15
Disable toggleRevealedButton (fixes #1243)

### LINKED ISSUE #1243: disable hypo toggle button if revealed card is clued, played, discarded
@padiwik

padiwik:


client/src/game/ui/Button2.ts
```diff
@@ -112,4 +114,22 @@ export default class Button2 extends Konva.Group {
   setMiddleText(text: string) {
     this.textElement2.text(text);
   }
+
+  setEnabled(enabled: boolean) {
```
robertos, line 118: if you do this, enable must become a readonly field. Otherwise someone can change it outside of this method. You probably want to use a proper getter and setter

padiwik: you mean the enabled property should be private?

robertos:
```
class Test
{
    private _prop: string;
    public get prop() : string
    {
        return this._prop;
    }

    doSomething() {
        this._prop = 'I can set it!';
    }
}
```

client/src/game/ui/hypothetical.ts
```diff
@@ -46,6 +46,8 @@ export const start = () => {
     replay.toggleSharedTurns();
   }
 
+  globals.elements.toggleRevealedButton?.setEnabled(true);
```
Zamiell, line 49: should be a ! instead of a ? i think

client/src/game/ui/hypothetical.ts
```diff
@@ -76,11 +78,13 @@ export const playThroughPastActions = () => {
   // If we are joining a hypothetical that is already in progress
   // or we are going backwards in an existing hypothetical,
   // play all of the existing hypothetical actions that have taken place so far, if any
+  globals.elements.toggleRevealedButton?.setEnabled(true);
```
Zamiell, line 81: same here

client/src/game/ui/hypothetical.ts
```diff
@@ -404,36 +408,35 @@ const cycleHand = () => {
...
 };
+
+// Check if we need to disable the toggleRevealedButton
+// This happens when a newly drawn card is played, discarded, or clued
+// It should also happen for misplays once those are implemented
+export const checkToggleRevealedButton = (actionMessage: Action) => {
+  if (actionMessage.type === 'play' || actionMessage.type === 'discard') {
+    const cardOrder = actionMessage.which.order;
+    if (globals.hypoFirstDrawnIndex && cardOrder >= globals.hypoFirstDrawnIndex) {
+      globals.elements.toggleRevealedButton?.setEnabled(false);
```
Zamiell, line 432: here

client/src/game/ui/hypothetical.ts
```diff
@@ -404,36 +408,35 @@ const cycleHand = () => {
...
+    }
+  } else if (actionMessage.type === 'clue') {
+    for (const cardOrder of actionMessage.list) {
+      if (globals.hypoFirstDrawnIndex && cardOrder >= globals.hypoFirstDrawnIndex) {
+        globals.elements.toggleRevealedButton?.setEnabled(false);
```
Zamiell, line 437: here

client/src/game/ui/hypothetical.ts
padiwik: we want to error out if the button doesn't exist for whatever reason?

Zamiell: i guess not

robertos: if we want to error out, why does the type allow null/undefined?

Zamiell: because its always supposed to exist but you can't instantiate it in the constructor because typescript is dumb

PR #1248, opened 2020-06-15
Simplify Button2 (extends Button), rename to ThreeLineButton

robertos:
Change text from "Toggle Hidden/Revealed Cards" to "Show/Hide Drawn Cards"


client/src/game/ui/ThreeLineButton.ts
```diff
@@ -0,0 +1,74 @@
...
+  textElement1: Konva.Text;
+  textElement2: Konva.Text;
+  textElement3: Konva.Text;
+
+  constructor(config: Konva.ContainerConfig) {
+    super(config);
+    this.listening(true);
+
+    // Hide the default label
+    this.textElement?.hide();
```
Zamiell, line 16: should this be ! instead of ?

client/src/game/ui/ThreeLineButton.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.add(this.textElement2);
+
+    this.textElement3 = this.textElement1?.clone({
+      y: 0.65 * h,
+      text: config.text3 as string | undefined,
+    }) as FitText;
+    this.add(this.textElement3);
+  }
+
+  setText(lines: {line1?: string, line2?: string, line3?: string}) {
```
Zamiell, line 50: why not just have 3 args

client/src/game/ui/ThreeLineButton.ts
robertos: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining
Means call hide() if not null, don't call anything otherwise

Zamiell: k

client/src/game/ui/ThreeLineButton.ts
robertos: because it makes it impossible to change line 2 without changing line 1 as well or writing ugly code like
setText(null, null, 'some text')

Zamiell: i see

client/src/game/ui/ThreeLineButton.ts
robertos: It's actually the opposite of ! :)

Issue #1219, opened 2020-06-10
Efficiency in clue starved variant


TheDaniMan:
Whenever a card bombs, the in-game efficiency counter increases the denominator by 1 (as discarding it would give back a clue, so effectively bombing wastes a clue).
In clue starved, it also seems to increase the denominator by 1; it should do something sensible, like increase it by 0.5, or better yet by 0/1 depending on the total number of clues: if the total number of clues (including discards and all but one 5) is an integer, increase the denominator for the first bomb but not the second, and if it is a half-integer increase it by 1 for the second bomb but not the first.

padiwik:
Are you also suggesting that discards should only update efficiency if it gives an actual full clue back?

Zamiell:
should i close this

robertos:
I believe so. Double check on a clue starved replay?

Zamiell:
thanks dani, this will be fixed in an upcoming DLC

Zamiell:
tests for this are in gameStateReducer.test.ts

PR #1251, opened 2020-06-15
Fix clue starved strike efficiency

padiwik:


client/src/game/ui/action.ts
```diff
@@ -508,7 +508,8 @@ actionFunctions.set('strike', (data: ActionStrike) => {
   const strikeX = globals.elements.strikeXs[i];
 
   // Update the stats
-  globals.cluesSpentPlusStrikes += 1;
+  // In clue starved variants, each strike only "costs" half a clue
+  globals.cluesSpentPlusStrikes += globals.variant.name.startsWith('Clue Starved') ? 0.5 : 1;
```
robertos, line 512: variantRules.isClueStaved(variant)
Let's not add more string based comparisons now that there is a better way

robertos:
Fix #1219 
Let's try to comment on the related issues so that 1. No need to close manually and 2. The issue opener or other people can easily check the fix

client/src/game/ui/action.ts
padiwik: is that implemented already? i would expect the rest of the code to be refactored to use this

Zamiell: variants.ts line 7

robertos: I'll log an issue to replace this kind of thing everywhere

PR #1258, opened 2020-06-16
Improve replay loading (fix #1254)

### LINKED ISSUE #1254: Going back in a replay doesn't always update cards
Sometimes cards that should be faded are not faded. And critical indicators aren't there when they should be.

padiwik:
After the last action is done when playing through past actions, we should always do a cardStatusCheck()

The deleted code in websocket.ts is redundant (and causes lag) because it already happens in the replay.goto function

### DIFF
--- a/client/src/game/ui/websocket.ts
+++ b/client/src/game/ui/websocket.ts
@@ -12,7 +12,6 @@ import ReplayArrowOrder from '../types/ReplayArrowOrder';
 import State from '../types/State';
 import action from './action';
 import * as arrows from './arrows';
-import cardStatusCheck from './cardStatusCheck';
 import { checkLegal } from './clues';
 import globals from './globals';
 import * as hypothetical from './hypothetical';
@@ -487,9 +486,6 @@ commands.set('gameActionList', (data: GameActionListData) => {
     replay.goto(turnNum, true);
   }
 
-  cardStatusCheck();
-  globals.layers.card.batchDraw();
-  globals.layers.UI.batchDraw();
   globals.loading = false;
 });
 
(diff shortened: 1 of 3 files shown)


client/src/game/ui/websocket.ts
```diff
@@ -487,9 +486,6 @@ commands.set('gameActionList', (data: GameActionListData) => {
     replay.goto(turnNum, true);
   }
 
-  cardStatusCheck();
```
robertos, line 490: I have a feeling this might break the non-replay case

padiwik: Oh this might break shared replays, let me check

padiwik: idk it seems to work, which case do you mean?

robertos: play a game, then in the middle of the game refresh one of the users

padiwik: seems fine

robertos: So who calls the batchdraw functions in that case? What I'm trying to avoid is, we're "getting lucky" in a case that doesn't require the redraw.

padiwik: Ugh I understand but it works in all the cases i can think of. Is it really the job of the function that plays through all the turns to reload the UI? Sidenote, these lines are very old see [the blame](https://github.com/Zamiell/hanabi-live/blame/5298753c9677ca24027f844d252ce60c2ed74521/public/js/src/game/ui/websocket.js)

Zamiell:
keep this PR in mind in case we see new UI bugs

PR #1259, opened 2020-06-16
Use Redux to manage state

robertos:
Finally we got to the point where the state can be managed by a Redux store.
Move the state change code from Lodash (deep clone) to Immer (partial clones using mutable drafts), for simplicity and likely performance
Readonly all the things


Zamiell:
thank you!!

Issue #1257, opened 2020-06-16
Avoid showing restarted games in the lobby


padiwik:
These games appear briefly in the lobby, fill up, and get instantly started. This is bad. Also the players briefly see the lobby, which is confusing. And these games can notify people on the waiting list, which they certainly shouldn't.

padiwik:
This is fixed except for the part where the players briefly see the lobby. That's a separate issue and should probably be fixed by making sure the loading screen shows properly in all cases

PR #1261, opened 2020-06-16
Improve restarted tables in the lobby

padiwik:
* Now a restarted table matches the original table name with a number appended
* These games should no longer appear in the main lobby while they are filling up (see #1257 )

### DIFF
--- a/server/src/command_table_create.go
+++ b/server/src/command_table_create.go
@@ -364,7 +364,7 @@ func createTable(s *Session, d *CommandData, preGameVisible bool) {
 	}
 
 	t := NewTable(d.Name, s.UserID())
-	t.PreGameVisible = preGameVisible
+	t.Visible = preGameVisible
 	t.PasswordHash = passwordHash
 	t.AlertWaiters = d.AlertWaiters
 	if setReplayOptions == nil {
--- a/server/src/command_table_create.go
+++ b/server/src/command_table_create.go
@@ -70,7 +70,7 @@ func commandTableCreate(s *Session, d *CommandData) {
 
 	// Make a default game name if they did not provide one
 	if len(d.Name) == 0 {
-		d.Name = s.Username() + "'s game"
+		d.Name = getName()
 	}
 
 	// Check for non-ASCII characters
@@ -100,6 +100,13 @@ func commandTableCreate(s *Session, d *CommandData) {
 		return
 	}
 
+	createTable(s, d, true)
+}
+
+// This function is run after some validation in commandTableCreate
+// that may be bypassed if the server creates the game from a restart, for example
+// preGameVisible is false if this game should be hidden before it starts, such as a restarted game
+func createTable(s *Session, d *CommandData, preGameVisible bool) {
 	// Set default values for the custom game options
 	var customDeck []SimpleCard
 	setSeedSuffix := ""
@@ -357,6 +364,7 @@ func commandTableCreate(s *Session, d *CommandData) {
 	}
 
 	t := NewTable(d.Name, s.UserID())
+	t.PreGameVisible = preGameVisible
 	t.PasswordHash = passwordHash
 	t.AlertWaiters = d.AlertWaiters
 	if setReplayOptions == nil {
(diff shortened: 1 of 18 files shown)


server/src/command_table_create.go
```diff
@@ -357,6 +364,7 @@ func commandTableCreate(s *Session, d *CommandData) {
 	}
 
 	t := NewTable(d.Name, s.UserID())
+	t.PreGameVisible = preGameVisible
```
Zamiell, line 367: we don't need to add a new property here, do we?
just use the existing t.Visible, and then set it to true after the start starts in the command_table_restart function

server/src/command_table_create.go
```diff
@@ -382,7 +390,8 @@ func commandTableCreate(s *Session, d *CommandData) {
 	// Alert the people on the waiting list, if any
 	// (even if they check the "Alert people on the waiting list" checkbox,
 	// we don't want to alert on password-protected games or test games)
-	if t.AlertWaiters &&
+	if preGameVisible &&
```
Zamiell, line 393: afaik this condition will never be met, so it is probably pointless

server/src/command_table_create.go
padiwik: oh yeah that should work, will do. but the function should still keep the preGameVisible argument right?

Zamiell: yes because you need to emulate whats happening on command_replay_create.go line 84

server/src/command_table_restart.go
```diff
@@ -1,11 +1,17 @@
...
 
 import (
+	"regexp"
 	"strconv"
 	"strings"
 )
 
-// commandTableRestart is sent when the user is in a shared replay of a speedrun game and wants to
+var (
+	// In example string "room name (#2)" this matches "room name" and "2"
```
Zamiell, line 10: could be less verbose
`// e.g. "room name (#2)" matches "room name" and "2"`

server/src/command_table_restart.go
```diff
@@ -152,16 +158,31 @@ func commandTableRestart(s *Session, d *CommandData) {
 		})
 	}
 
-	// Generate a random name for the new game
-	newTableName := getName()
+	newTableName := ""
+
+	// Generate a new name for the game based on which iteration of the room this is
+	// For example, a game named "logic only" will be "logic only (#2)" and then "logic only (#3)"
+	if t.InitialName != "" {
```
Zamiell, line 165: i dont think it is necessary to add a new field here, it strikes me as cluttering the table oject.
rather, can't we infer the initial name by using regex in e.g. the exact same way that you increment the number?

server/src/command_table_create.go
padiwik: No, in the normal case (user creates a table) preGameVisible is true

Zamiell: allow me to clarify. by design, the waiting list will always be empty when the code reaches this point. thus, your added check doesn't do anything really

server/src/command_table_restart.go
padiwik: No because then the game would be "Shared Replay for game #N" 

and I don't think old table structures are kept

Zamiell: oh, i see, that is unfortunate. maybe we should just get rid of table names altogether

padiwik: Hyphenated doesn't use table names but the majority of non-hyphenated does

Zamiell: "using" a table name isn't very specific. they might make a custom table name with a silly, arbitrary string, but that isn't actually adding any value to anybody. on the contrary, with the name column removed from the main lobby screen, you have much less clutter

padiwik: still needed for e.g. don't accidentally spectate a competition game, unless there's a big refactor in how the lobby is drawn (hide these games?).
also i wouldn't recommend disabling table names until people who don't use chat commands know they can friend people

Zamiell: agreed, maybe something to think about in the future

padiwik:
Ok it seems the warning that occurs when someone is on the waiting list never even happens, i should fix that

Zamiell:
thanks

PR #1262, opened 2020-06-16
Improve switching the leader

### LINKED ISSUE #1260: Passing leader during a hypo breaks the interface
broken interface

padiwik:
* No longer need to right click the crown for it to work, any old click will do
* Fix #1260

### DIFF
--- a/client/src/chatCommands.ts
+++ b/client/src/chatCommands.ts
@@ -109,7 +109,7 @@ const setLeader = (_room: string, args: string[]) => {
 chatCommands.set('setleader', setLeader);
 chatCommands.set('setlead', setLeader);
 chatCommands.set('setowner', setLeader);
-chatCommands.set('chnageleader', setLeader);
+chatCommands.set('changeleader', setLeader);
 chatCommands.set('changelead', setLeader);
 chatCommands.set('changeowner', setLeader);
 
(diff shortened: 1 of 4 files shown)


client/src/chatCommands.ts
```diff
@@ -109,7 +109,7 @@ const setLeader = (_room: string, args: string[]) => {
 chatCommands.set('setleader', setLeader);
```
robertos, line 109: Since you're in the area, might as well end a hypo if in progress? 
#1260

robertos: Oh seems like you took care below. Nevermind

client/src/game/ui/websocket.ts
```diff
@@ -598,10 +598,11 @@ commands.set('replayLeader', (data: ReplayLeaderData) => {
   // Hide the replay area if we are in a hypothetical
   if (globals.hypothetical) {
     hypothetical.show();
+  } else {
+    // A batchDraw() already happens in the hypothetical.show() function
```
Zamiell, line 602: the point of batch draw is that you can call it N times and it will incur no additional cost. (as opposed to the normal draw function which will actually always redraw the layer instantly)

thus its okay to just call batch draw judiciously whenever you need it

client/src/chatCommands.ts
padiwik: Intentionally not done. You can pass lead to someone else and have them continue your hypo, it works fine

client/src/game/ui/websocket.ts
padiwik: Oh that's good to know, I won't spend time trying to kill them in the future

Zamiell:
thanks

Issue #1264, opened 2020-06-17
Up or Down: s on notes should denote start cards instead of 7 (e.g. sg = start of green)


robertos:

Zamiell:
conventionally we use r0, s0, etc.
but that doesn't work either currently so yeah this issue needs to remain open

padiwik:
Is there any suit that has the "s" abbreviation?

robertos:
We can use 0 if there is/was/could be

Zamiell:
sky, sapphire

padiwik:
so the ambiguous suits, and those will most likely not be up or down

Zamiell:
for reference, it looks like padi is currently working on this

Zamiell:
seems to be fixed now thanks to padi

PR #1267, opened 2020-06-17
Better notes

padiwik:
See also #1264

### DIFF
--- a/client/src/game/ui/notes.ts
+++ b/client/src/game/ui/notes.ts
@@ -1,7 +1,10 @@
 // Users can right-click cards to record information on them
 
 // Imports
+import * as variantRules from '../rules/variant';
+import { CardIdentity } from '../types/CardIdentity';
 import CardState from '../types/CardState';
+import { START_CARD_RANK } from '../types/constants';
 import globals from './globals';
 import HanabiCard from './HanabiCard';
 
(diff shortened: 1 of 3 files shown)


client/src/game/ui/notes.ts
```diff
@@ -104,126 +107,100 @@ export const checkSpecialNote = (card: HanabiCard) => {
...
-  );
+  card.state.noteKnownTrash = checkNoteKeywords([
+    'kt',
+    'trash',
+    'stale',
+    'bad',
+  ], note, fullNote);
+
+  card.state.noteNeedsFix = checkNoteKeywords([
+    'fixme',
```
Zamiell, line 126: small thing - make this one line, and if we add an alias we can split it into separate lines

padiwik: ok, i was wondering about those

client/src/game/ui/notes.ts
```diff
@@ -104,126 +107,100 @@ export const checkSpecialNote = (card: HanabiCard) => {
...
+  ], note, fullNote);
+
+  card.state.noteFinessed = checkNoteKeywords([
+    'f',
+    'hf',
+    'pf',
+    'gd',
+  ], note, fullNote);
+
+  card.state.noteBlank = checkNoteKeywords([
```
Zamiell, line 151: same here

client/src/game/ui/notes.ts
```diff
@@ -104,126 +107,100 @@ export const checkSpecialNote = (card: HanabiCard) => {
...
+    'blank',
+  ], note, fullNote);
+
+  card.state.noteUnclued = checkNoteKeywords([
+    'unclued',
```
Zamiell, line 156: and here

Zamiell:
thank you!!

PR #1270, opened 2020-06-18
Going to the lobby now resets the URL

padiwik:
Usecase: I opened hanabi.live via a /replay link in discord. I press go to lobby. I don't want to view this replay anymore every time I refresh the page


padiwik:
Closing because it breaks endpoints like `/dev` and `/test`. See #1271 which should resolve this issue when it's implemented

PR #1281, opened 2020-06-19
Fix /changevariant user stats

### LINKED ISSUE #1280: Same stats for different players
image

padiwik:
Fixes #1280


Zamiell:
nice catch

PR #1289, opened 2020-06-20
Start changing UI using states!

robertos:


Zamiell:
time to implement some views ><

PR #1292, opened 2020-06-20
Touch actions!

padiwik:


client/src/game/ui/HanabiCardClick.ts
```diff
@@ -79,7 +79,7 @@ const clickMiddle = (card: HanabiCard, event: MouseEvent) => {
     return;
   }
 
-  // Middle clicking on a card goes to a turn it was first clued
+  // Middle clicking on a card goes to the turn it was first clued
```
Zamiell, line 82: nice find ><

client/src/game/ui/drawUI.ts
```diff
@@ -1395,15 +1399,13 @@ const drawTimers = () => {
     listening: true,
   });
   globals.layers.timer.add(globals.elements.timer1 as any);
-  globals.elements.timer1.on('click', (event) => {
+  function timerClick() {
```
Zamiell, line 1402: always prefer `const timerClick = () => {` to `function timerClick() {`  UNLESS you need `this`, and it doesn't look like you are using `this` here

client/src/misc.ts
```diff
@@ -2,6 +2,8 @@
 // NOTE: This is because this file introduces a generic extension for
 // all JQuery objects
 
+import globals from './game/ui/globals';
```
Zamiell, line 5: misc probably shouldn't be importing globals. how about we move `resetActiveHover()` to `client/src/game/ui/tooltips.ts`

padiwik:
k I integrated your feedback and rebased to master

client/src/game/ui/HanabiCard.ts
```diff
@@ -34,6 +34,8 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   bareName: string = '';
 
   tweening: boolean = false;
+  wasRecentlyTapped: boolean = false;
+  touchstartTimeout?: NodeJS.Timeout | null;
```
Zamiell, line 38: 1) type is wrong on this https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript
2) pretty sure this needs to get reset on refresh

Zamiell:
mobile support! (ღ˘˘ღ)

padiwik:
guaranteed to have bugs, you saw how it takes two extra variables just to get tapping and double-tapping to work

Zamiell:
as long as the bugs dont occur for non-mobile users, its better than it was

PR #1297, opened 2020-06-21
Add hypothetical support to reducers

robertos:
+ separate replay processing in its own reducer
+ basic tests for hypothetical


Zamiell:
thank you!!

PR #1298, opened 2020-06-21
Add pace/risk updates from the state observer

robertos:


Zamiell:
nice, i was gonna do this next but u beat me to it =p

PR #1299, opened 2020-06-21
Move controls to ui/controls

robertos:
A control is a UI type that doesn't depend on any application types or logic to render. Good examples are buttons, shapes, or text boxes.

+ Make some interfaces export default
+ Sub LearnedCard with CardIdentity since revealed was not used
+ Move SpectatorNote to types
+ No functionality changes


Zamiell:
all hail controls

PR #1305, opened 2020-06-21
Revert scaling changes to match old UI on a 1080p screen

padiwik:
Specifically, zam's canvas which is 947 pixels high


Zamiell:
thank you!

PR #1310, opened 2020-06-22
Split card.reveal() functionality into convert()

padiwik:
Also should fix the issue where for example in this review http://hanabi.live/replay/214544/24 the g3 is marked as not 4 and not 1 because pablo himself drew g4 and g1.


Zamiell:
thanks

Issue #1309, opened 2020-06-21
make "Hanabi Live" text disappear when overlapping with icons on mobile


Zamiell:
@padiwik

padiwik:
Also make it link to the lobby as rob requested

Zamiell:
fixed thanks to padi

PR #1313, opened 2020-06-22
Better mobile nav bar; Hanabi Live is now a link in header

padiwik:
See #1309

### DIFF
--- a/server/src/views/main.tmpl
+++ b/server/src/views/main.tmpl
@@ -4,7 +4,7 @@
 
   <!-- Header -->
   <header id="header">
-    <h1><a href="/">Hanabi Live</a></h1>
+    <h1>Hanabi Live</h1>
     <nav id="nav">
       <!-- We use <i> instead of the icon class because it results in a more compact button -->
       <ul id="nav-buttons-games" class="hidden">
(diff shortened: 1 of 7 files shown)


server/src/views/main.tmpl
```diff
@@ -4,7 +4,7 @@
 
   <!-- Header -->
   <header id="header">
-    <h1>Hanabi Live</h1>
+    <h1><a href="/">Hanabi Live</a></h1>
```
Zamiell, line 7: it should't be on main because that will fuck up the user

padiwik: why? it's just a reload

Zamiell: if the user is on a slow computer or a slow internet connection, one stray misclick will cause a frustrating 30 second period where they cannot do anything anymore. that is terrible

Zamiell:
confirmed on both main and other sites, looks good

Zamiell:
padi i think the easiest way is to just make a breakpoint that removes the padding on the buttons

padiwik:
did u mean to comment on valetta's issue?

Zamiell:
yes oops

PR #1320, opened 2020-06-23
Better profile buttons

### LINKED ISSUE #1314: tab is outside the border
49305DEA-DD8F-4E61-874B-FC0852C620A1

padiwik:
Fixes #1314 . They're all stacked on top of each other because I think it would be weird to have two rows of two buttons each in this layout.


Zamiell:
fantastic

PR #1323, opened 2020-06-24
Make Options readonly

robertos:


Zamiell:
this guy is an es6 wizard

Issue #1322, opened 2020-06-23
Joining a hypo in progress gives errors on new hypoActions


padiwik:
```
Uncaught Error: [Immer] minified error nr: 23 . Find the full error at: https://bit.ly/3cXEKWf
    at n (immer.esm.js:1)
    at o (immer.esm.js:1)
    at replayReducer.ts:44
    at immer.esm.js:1
    at t.e.produce (immer.esm.js:1)
    at immer.esm.js:1
    at stateReducer.ts:43
    at immer.esm.js:1
    at t.e.produce (immer.esm.js:1)
    at immer.esm.js:1
    at Object.v [as dispatch] (redux.js:213)
    at websocket.ts:151
    at Object.hypoAction (websocketInit.ts:29)
    at n.onMessage (Connection.ts:52)```

robertos:
Original expects a draft, got something else

robertos:
Likely some undefined

Zamiell:
fixed thanks to rob!

PR #1325, opened 2020-06-24
Fix joining a hypothetical in progress

robertos:


robertos:
Fix #1322

Zamiell:
thanks as always!

PR #1327, opened 2020-06-24
Don't hide tooltips when a game action happens

padiwik:
Suppose Bob plays. If you're reading a note on Cathy's hand, you want to keep reading it. If you're reading a note on Bob's hand and the card moves, the note will automatically close when you move your mouse a bit.

### DIFF
--- a/client/src/game/ui/action.ts
+++ b/client/src/game/ui/action.ts
@@ -32,7 +32,6 @@ import HanabiCard from './HanabiCard';
 import LayoutChild from './LayoutChild';
 import possibilitiesCheck from './possibilitiesCheck';
 import strikeRecord from './strikeRecord';
-import * as tooltips from './tooltips';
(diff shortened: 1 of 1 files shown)


padiwik:
It has existed since initial commit https://github.com/Zamiell/hanabi-live/blame/9b2a50c21d29131fe27a952b97e06ddba7b51388/public/js/ui.js

Zamiell:
this code is intentional, because it fixes the "bug" where:
1) Alice is hovering over Bob's slot 1 and reading the note that she has previously written there
2) Bob plays his slot 1 card, and the card moves from his hand to the play stacks
3) the note tooltip stays in place and is now hovered over the wrong card, the card that Bob just drew

padiwik:
This is not an issue. If Alice wants to stop reading the note, she just needs to move her mouse a bit and it goes away.

Zamiell:
ok then

PR #1333, opened 2020-06-24
Turn off empathy on played/discarded cards

padiwik:
Reinstate #1327 , but fix the glitch where played/discarded cards continue to appear hidden due to empathy


Zamiell:
cool, thanks

PR #1342, opened 2020-06-26
Hide special borders on played/discarded cards

### LINKED ISSUE #1311: Don't draw special borders on played/etc cards after reloading

padiwik:
Fixes #1311


Zamiell:
thank you

Issue #1351, opened 2020-06-28
ending hypothetical doesn't reset morphed cards anymore, probably because of recent changes


Zamiell:

padiwik:
like how recent? cuz it might be from me functionizing the card.reveal() function

padiwik:
it also doesnt unblank cards, but going into empathy fixes it. so maybe it is the fault of not calling setBareImage()

padiwik:
yeah cuz when you're following a shared replay, people can draw completely blank cards. so it seems like `globals.turn` is mismatched or something. it's probably a different value for the shared turn as opposed to the local turn shuttle.

Zamiell:
also we found a bug where cards were blanked out after we ended the hypo, which is probably just a different manifestation of the same bug

padiwik:
also toggling hypoRevealed doesn't redraw the cards anymore

Zamiell:
padi should i close this now?

padiwik:
sure, ill submit the other PR as well one sec

padiwik:
ok, don't close this. at least two issues still exist:

* doesn't reset morphed cards on hypo end
* enter a replay, go to some intermediate turn, hold empathy on a card. then go back a turn. it seems to blank out some of the card information

Both of which are caused by the `setBareImage()` optimization and can be fixed by toggling empathy. But I can't figure out a good fix.

padiwik:
I've got the fix for the second bug with calling `resetActiveHover()`. But what's the best way to fix the first bug? Either we need to redraw all the cards drawn up to the current turn, which seems inefficient, or keep track of which cards were morphed and redraw all of them.

padiwik:
Oh yknow, the first version shouldn't be any less efficient because they need to be drawn anyways

Zamiell:
i vote first version

Zamiell:
closing now, seems to be fixed on card-reducer too

PR #1355, opened 2020-06-28
Bugfix for setBareImage() optimization

padiwik:
Should fix the blank cards described in #1351 . I don't really think there's a more efficient way to do this without restructuring everything.

### DIFF
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -201,6 +201,7 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
     if (
       this.bareName !== ''
       && globals.replay
+      && !globals.hypothetical
       && globals.turn < globals.replayTurn - 1
     ) {
       return;
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -196,10 +196,12 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   setBareImage() {
     // Optimization: This function is expensive, so don't do it in replays
     // unless we got to the final destination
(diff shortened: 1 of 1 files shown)


padiwik:
Hmmm it's still bugged in the case where you're holding empathy and go back a turn

padiwik:
Instead, I think we need to resetActiveHover() when an action happens in a shared replay, since this performance improvement is better than being able to actually read notes when the leader is scrolling quickly through a shared replay. If you really want to view empathy/read notes, you could Pause Shared Turns for a bit.

Zamiell:
can you submit a pr for that?

client/src/game/ui/HanabiCard.ts
```diff
@@ -196,10 +196,13 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   setBareImage() {
     // Optimization: This function is expensive, so don't do it in replays
     // unless we got to the final destination
+    // However, if an action happens before the "turn" message is sent,
+    // we still need to redraw any affected cards
     if (
       this.bareName !== ''
       && globals.replay
-      && globals.replayTurn !== globals.turn
+      && !globals.hypothetical
```
padiwik, line 204: oh this is actually unnecessary if we use the inequality, since globals.replayTurn will always be less than or equal to globals.turn in a hypo

Zamiell: not necessarily, if we are on the 0th turn of a hypo

padiwik: at the start of a hypo, they are equal, aren't they?

Zamiell: im not sure, you could test

PR #1363, opened 2020-06-30
Fix state turn going out of sync on game over

### LINKED ISSUE #1343: Score is always 0/30 at endgame
Score shows as 0/30 at game finish regardless of actual score. This has been happening for ~1 week on every game played. 

Example Game ID: 219037

image

robertos:
Fix #1343


Zamiell:
thanks

PR #1364, opened 2020-06-30
Fix side effects of setBareImage() optimization

padiwik:


padiwik:
Fix #1351

Zamiell:
Padi the bug slayer

padiwik:
i can't keep up with rob! :D

PR #1365, opened 2020-06-30
Add starting turn to startReplay action

### LINKED ISSUE #1362: Entering in-game replay doesn't load the stats correctly
Pace will be 1 value too high and the efficiency will be blank. Also, the current player indicator (bolded nameframe) might be wrong or empty. This can be fixed by going back and forth in the replay

robertos:
Fix #1362


Zamiell:
so startReplay needs to always be given a turn?

PR #1366, opened 2020-07-01
Pull from master, more cleanup on applyClueCore

robertos:


spell_check/spell_check.sh
```diff
@@ -17,6 +17,7 @@ FILES=$(find "$DIR" -type f \
   -not -path "$DIR/client/test_data/*" \
   -not -path "$DIR/client/webpack_output/*" \
   -not -path "$DIR/client/package-lock.json" \
+  -not -path "$DIR/client/test_data/*" \
```
Zamiell, line 20: dupe line

robertos: Sure

PR #1367, opened 2020-07-01
Create cardPossibilitiesReducer (aka applyClue)

robertos:
Add a whole set of tests to ensure everything is working fine before merging.
Still missing some logic for updating possibilities on other cards, revealing, etc.
Spoiler: possibility tests are not yet passing (skipped)


client/test/loadGameJSON.ts
```diff
@@ -107,7 +107,9 @@ export default function loadGameJSON(gameJSON: JSONGame): State {
         // Check if this is actually a play or a misplay
         const jsonCard = gameJSON.deck[a.which.order];
         const playStack = s.playStacks[jsonCard.suit];
-        const topOfStackRank = playStack.length === 0 ? 0 : gameJSON.deck[playStack[playStack.length - 1]].rank;
+        const topOfStackRank = playStack.length === 0
```
Zamiell, line 111: this violates airbnb 15.6.
i think there also might be other places in the code where you have done this, not 100% sure

robertos: Not quite a violation but sure I can add another local

PR #1371, opened 2020-07-04
Reactive card properties!

robertos:
+ Enable linter to catch changes to params, more tests, merge master


client/src/Connection.ts
```diff
@@ -88,4 +85,4 @@ const unpack = (data: string) => {
   return [name, data.substring(name.length + 1, data.length)];
 };
 const unmarshal = (data: string) => JSON.parse(data) as unknown;
-const marshalAndPack = (name: string, data: string) => name + separator + JSON.stringify(data);
+const marshalAndPack = (name: string, data: any) => name + separator + JSON.stringify(data);
```
Zamiell, line 88: should it be an object with any properties instead of any?

robertos: i mean, it is any because JSON.stringify takes any. The string annotation there was wrong.

PR #1374, opened 2020-07-04
Revert #1372, use relative scaling instead

padiwik:


padiwik:
cuz https://cdn.discordapp.com/attachments/333680903748321282/729077270597075045/unknown.png happens if this code is deleted entirely

padiwik:
Actually it probably makes more sense to do it in terms of `uw` (card width) instead of absolute screen width. But this is the way that doesn't change any UI on zam's screen

PR #1394, opened 2020-07-07
Add deck animations and arrows. Reverse animations!

robertos:
+ Change location to be strongly typed
Includes all changes from #1392 , so that can be closed. @Zamiell you might want to take a look to see how I solved it before closing.


client/src/game/rules.ts
```diff
@@ -0,0 +1,11 @@
+// Index for all rules exports
```
Zamiell, line 1: is the point of this so that we can just do 1 rules import and then do `rules.card.poop()` ?

client/package.json
```diff
@@ -14,15 +14,16 @@
   "devDependencies": {
     "@sentry/webpack-plugin": "^1.11.1",
     "@types/jest": "^26.0.0",
-    "@typescript-eslint/eslint-plugin": "^3.3.0",
-    "@typescript-eslint/parser": "^3.3.0",
+    "@typescript-eslint/eslint-plugin": "^3.6.0",
```
Zamiell, line 17: make sure you always reinstall the linter after changing package.json.
eslint-plugin-import is too new of a version

client/src/game/rules.ts
robertos: No, it's so you can import rules like
import { cardRules, variantRules } from 'rules'
And VSCode will actually autocomplete the above and suggest names as you type.

PR #1395, opened 2020-07-07
End game early with reversed suits

### LINKED ISSUE #1193: bug in reverse code, doesn't know that k1 isn't playable on the final turn, so it extends the game longer than it needs to

padiwik:
Fix #1193 .... I had a cursory glance of the logic and it seemed correct, so this shouldn't have any negative side effects. It now matches `game.go` line 270


Zamiell:
did you test that it fixes by doing a !replay game?

padiwik:
i made a game from scratch, how do i do a valid !replay game when im in my testing environment lol

Zamiell:
one sec i'll update the live site

Issue #1403, opened 2020-07-08
card-reducer branch - stack bases disappear when rewinding


Zamiell:

Zamiell:
woot

Issue #1404, opened 2020-07-08
card-reducer branch - misplay animation doesn't work


Zamiell:

robertos:
That can likely be fixed on the view directly by checking the card's location/misplayed when it's removed from a hand

Zamiell:
yay

PR #1409, opened 2020-07-08
Add misplay animation, fix many animation issues

robertos:
+ Konva Tween helper ensuring fewer leaks and fewer mouse events


robertos:
Fix #1403 #1404

PR #1435, opened 2020-07-10
Bugfix for character metadata from db

padiwik:
Characters such as Mood Swings need to start off with -1 metadata, because otherwise server thinks invalid actions occur

### DIFF
--- a/server/src/character.go
+++ b/server/src/character.go
@@ -107,6 +107,11 @@ func characterGenerate(g *Game) {
 
 		for i, p := range dbPlayers {
 			g.Players[i].Character = charactersID[p.CharacterAssignment]
+			// Make sure to clear the character metadata for characters that don't depend on
+			// a random rank or color as read from the database
+			if p.CharacterAssignment > 3 {
+				p.CharacterMetadata = -1
+			}
 			g.Players[i].CharacterMetadata = p.CharacterMetadata
 		}
 		return


padiwik:
Perhaps these characters should also store 0 in the db to "save space", but would that cause any discernible improvement?

Zamiell:
hrm, i think this pr doesn't do anything, it is already initialized to -1 in character.go

```
	for i, p := range g.Players {
		// Initialize the metadata to -1 (it is 0 by default in order to save database space)
		p.CharacterMetadata = -1
```

padiwik:
False, that is done only when *creating* characters (running the RNG)

Zamiell:
oh

Zamiell:
i have an alternate solution
lets let charactermetadata be 0 when falsey, and a positive number when true

Zamiell:
its weird because then character metadata won't map directly to clue type, but i think thats okay, because `p.CharacterAssignment > 3 {` would break in the hypothetical where we added another character that needed to store metadata in the DB

padiwik:
that would require rewriting characters, consider for example the guys with [random color] = 0

padiwik:
which is fine, but a bit ugly

Zamiell:
right, we make a DB query:

if characterID < 4 & metadata = 0 {
  characterID = characterID + 1
}

padiwik:
oh i guess we can keep the numbers the same in the go code but shift by 1 on writing to or reading from db

Zamiell:
i want to rewrite the go code actually. i think using 0 and 1 makes more sense than using -1 and 0 like i was doing before

padiwik:
there'll be a lot of index shifting i think, but go ahead

padiwik:
will the typescript code also get shifted? etc.

Zamiell:
hrmmmm

Zamiell:
> oh i guess we can keep the numbers the same in the go code but shift by 1 on writing to or reading from db

ok lets go with this for now, since the client seems to have a convention of using -1 for null

Zamiell:
this means we need to write a database transformation for all Db metadata right?

padiwik:
that should be rather easy, am not worried about that

padiwik:
of course it needs to be run on any backup ugh

Zamiell:
ok, adding this code:

```
					// -1 is considered to be "null" metadata
					// Since most characters have null metadata,
					// we save space in the database by storing -1 as 0
					// However, if some characters have a metadata of 0, it is meaningful
					// (e.g. corresponding to the 0th suit)
					// Thus, we store meaningful metadata in the database as 1 + the value to make
					// it clear that it is not a null value
					if v.WriteMetadataToDatabase {
						characterMetadata = gp.CharacterMetadata + 1
					}
```

PR #1440, opened 2020-07-11
Fix hypotheticals / morphing

robertos:
+ Remove blank from CardState
+ Consolidate hypo state in a new type
+ Move touchedCard to cluesRules
+ Remove some unused globals
+ TODO: morphing stack bases doesn't work because stack bases are not in the deck


client/src/game/types/ReplayState.ts
```diff
@@ -1,9 +1,19 @@
...
 export default interface ReplayState {
   readonly active: boolean;
   readonly turn: number;
   readonly states: readonly GameState[];
-  readonly ongoingHypothetical: GameState | null;
-  readonly hypotheticalStates: readonly GameState[];
+  readonly hypothetical: HypotheticalState | null;
+}
+
+export interface HypotheticalState {
```
Zamiell, line 11: doesn't this belong in its own file

robertos: Potentially. one could argue we can make a hypo reducer as well... I'm going for parity first :)

Zamiell:
looks good, thanks

PR #1445, opened 2020-07-12
Fix #1444: disallow using nullables/undefineds as booleans (WAT)

### LINKED ISSUE #1444: card-reducer branch: morphing to red doesn't work

robertos:


client/src/game/ui/HanabiCard.ts
```diff
@@ -687,7 +687,7 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
 
   animateToDeck() {
     const layoutChild = this.parent as unknown as LayoutChild;
-    if (!layoutChild || !layoutChild.parent) {
+    if (layoutChild === undefined || !layoutChild.parent) {
```
Zamiell, line 690: should this be
if (layoutChild === undefined || layoutChild.parent === undefined) {

Zamiell: if (layoutChild === undefined || layoutChild === null || layoutChild.parent === undefined || layoutChild.parent === null) {

robertos: parent can be null or undefined, i think.  I can add another condition here, but the linter allowed this one...

client/src/game/data/suitsInit.ts
```diff
@@ -92,7 +92,7 @@ export default function suitsInit(COLORS: Map<string, Color>) {
           throw new Error(`The color "${colorString}" in the suit "${suitName}" does not exist.`);
         }
       }
-    } else if (!suitJSON.allClueColors && !suitJSON.noClueColors) {
+    } else if (suitJSON.allClueColors !== true && suitJSON.noClueColors !== true) {
```
Zamiell, line 95: lol, use === false

client/src/game/ui/CardLayout.ts
```diff
@@ -135,7 +135,7 @@ export default class CardLayout extends Konva.Group {
             opacity: 1,
             easing: Konva.Easings.EaseOut,
             onFinish: () => {
-              if (!card || !layoutChild) {
+              if (card === undefined || layoutChild === undefined) {
```
Zamiell, line 138: add both null checks

client/src/game/data/suitsInit.ts
robertos: NOPE: !== true => can be false or undefined. This is exactly the bug!!

client/src/game/ui/CardLayout.ts
robertos: those can't be null, sorry :)

client/src/game/data/suitsInit.ts
Zamiell: write a comment then explaining the code that this isn't a bug

client/src/game/ui/CardLayout.ts
robertos: their types are HanabiCard and Konva.Node, none of which is nullable.

client/src/game/ui/HanabiCard.ts
```diff
@@ -1054,14 +1058,14 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
         return;
       }
       const hand = this.parent.parent as unknown as CardLayout;
```
Zamiell, line 1060: null check here, and check for this.parent before assigning this.parent.parent (?)

client/src/game/ui/PlayStack.ts
```diff
@@ -44,7 +44,7 @@ export default class PlayStack extends Konva.Group {
         opacity,
         easing: Konva.Easings.EaseOut,
         onFinish: () => {
-          if (!layoutChild || !card || !card.parent) {
+          if (layoutChild === undefined || card === undefined || card.parent === undefined) {
```
Zamiell, line 47: null check here

client/src/game/ui/HanabiCard.ts
robertos: no need, hand will be undefined if any of those happen

client/src/game/ui/PlayStack.ts
robertos: nope, can't be null. you gotta check the types

Zamiell: card.parent can't be null?
then why do you check if it is null in other places

robertos: Because Konva, as you know, is very consistent. Parent can be null in some types, and can't in others.

Zamiell: should we just check for parent === null everywhere? it makes the code more consistent, and could make the code more resilient if we ever upgrade konva

robertos: That's the thing, it doesn't really, because TS will catch if you forget a null check.

Zamiell:
so what was the actual bug fix amidst all of the noise?

robertos:
> so what was the actual bug fix amidst all of the noise?

HanabiCardClick 
  // We want an exact match, so fullNote is sent as an empty string
  const cardIdentity = notes.cardIdentityFromNote(globals.variant, cardText, '');
  if (!cardIdentity.suitIndex || !cardIdentity.rank) {
  if (cardIdentity.suitIndex === null || cardIdentity.rank === null) {
    window.alert('You entered an invalid card.');
    return;
  }

Issue #1446, opened 2020-07-12
card-reducer branch - in a reply, cards can be removed twice


asaelr:
steps to reproduce (for example):
Watch the replay in the attached file
Go to turn 4
Check the empathy for the clued p4
It has an X over the '2' pip, even though only a single p2 is seen.

I'm pretty sure that the reason for that is line 114 of cardsReducer.ts , combined with the fact that in a replay, _ourPlayerIndex_ is a valid player index.
[buggyGameJSON.txt](https://github.com/Zamiell/hanabi-live/files/4908168/buggyGameJSON.txt)

robertos:
Thanks Asael, does it repro on live?

robertos:
Also I think it might have been recently introduced by Zamiel adding ourPlayerIndex...

robertos:
I.e. #1439

asaelr:
> Thanks Asael, does it repro on live?

no. only on card-reducer.

robertos:
Should be fixed by #1452

Zamiell:
closing then, thanks rob

PR #1449, opened 2020-07-12
Card reducer

asaelr:
I rewrote the applyClue mechanism.
Basically, instead of tracking ranks and colors separately, every cardState has now a 2D boolean array, that tracks the cards that match to its clues.
(To simplify some calculations, it has also an array of matching cards)
A variant has now a pre-calculated arrays contains the "which clue touch which cards" data. Using them, the cardPossibilitiesReducer becomes much more simple. (We don't need the complicated logic of applyClueCore any more)


client/src/game/types/CardState.ts
```diff
@@ -9,10 +9,14 @@ export default interface CardState {
...
 
-  // possibleCards[suitIndex][rank] = how many cards of this suitIndex and rank could this be?
+  // matching[suitIndex][rank] = does the card of this suitIndex and rank match our clues?
+  // unseenCards[suitIndex][rank] = how many cards of this suitIndex and rank could this be?
   // NOTE: we're using an array as a map, so there will be empty spaces for ranks
   // that are not valid card ranks (e.g. 0, or 6 in Up or Down)
-  readonly possibleCards: ReadonlyArray<readonly number[]>;
+  readonly matchingCards: ReadonlyArray<readonly boolean[]>;
+  readonly unseenCards: ReadonlyArray<readonly number[]>;
+  readonly matchingCardsArray: ReadonlyArray<readonly [number, number]>;
```
robertos, line 18: matchingCardsArray is not a good name for this (since matchingCards is also an array...). It is a reverse index of matching cards, if i understand correctly?

client/src/game/types/CardState.ts
```diff
@@ -9,10 +9,14 @@ export default interface CardState {
   readonly rankClueMemory: ClueMemory;
   readonly colorClueMemory: ClueMemory;
 
-  // possibleCards[suitIndex][rank] = how many cards of this suitIndex and rank could this be?
+  // matching[suitIndex][rank] = does the card of this suitIndex and rank match our clues?
+  // unseenCards[suitIndex][rank] = how many cards of this suitIndex and rank could this be?
   // NOTE: we're using an array as a map, so there will be empty spaces for ranks
   // that are not valid card ranks (e.g. 0, or 6 in Up or Down)
-  readonly possibleCards: ReadonlyArray<readonly number[]>;
+  readonly matchingCards: ReadonlyArray<readonly boolean[]>;
```
robertos, line 16: not a fan of this name either. I see the comment, but what does "match our clues" mean? I.e. if i touch a pink with 2, any 2 and any pink "match" this clue. This map is stored on an array, so be careful with non-contiguous map/filter

client/test_data/up_or_down_turn5.json
```diff
@@ -103,7 +103,9 @@
         "turnDrawn": 0,
         "turnDiscarded": -1,
         "turnPlayed": -1,
-        "isMisplayed": false
+        "isMisplayed": false,
+        "matchingCards": [],
```
robertos, line 107: well, this invalidates our current tests :/ Is it possible to update the script under test_data to generate valid tests that match your rewrite?

asaelr:
Thanks for the review.

I did some of the work before the integration tests were skipped. They passed, but maybe I broke something since then...
I think that the only place where a problem can occur is with the "possibleCards" tests - I renamed it to "unseenCards", and excluded clue information from it (so it can have a positive number on y2 even if you have a negative yellow). I added (semi-manually) empty fields to the JSON files just to make it compile, but you are right that I should generate them in a correct way. I will check the script. (this addition doesn't invalidate the tests, as we don't test these fields)

I agree that pip states are not pure logic, but we probably want some logic variable with the exact values. Just like we have "possibilities", even though it can be derived from the "real" information variables (matchingCards and unseenCards).
Probably we want to rename the values (to something like 'Possible', 'Impossible from observation', 'Impossible by clues', but probably shorter), but that's just names.
(Of course, it's just the balance between how much information we save vs how complicated is the logic)

Probably rename matchingCardsArray to matchingCardsList? I'm not sure it will be more clear. (It's indeed the reverse index of matchingCards. I think that keeping both saves some time)
Indeed, the clue '2' match any 2 and any pink card. If you like another name, I can change.

About the non-contiguous: it's indeed error-prone, but it's the same format of unseenCards, and I didn't want to change all of it.
Ideally, parts of the code should use suit/rank, and different parts should use suitIndex/rankIndex. Now, many part use suitIndex/rank, which I find kinda confusing.

robertos:
I'll start with fixing the state of the world as it is now (with spectators) before i move to look at this more closely. I like the general approach but need to coalesce some things (e.g., do we still need the possibilities list?)

client/test_data/up_or_down_turn5.json
asaelr: Well I looked at that script. if I understood correctly, it just copies the data from the old logic. Since the new logic has more data (especially the separation between matchingCards and unseenCards), I don't think it's possible (without copy the whole algorithm).
Of course, I can change it so it will copy the data from the new logic, but then the tests won't verify that the behavior is the same as in the old code...
I think that observed behavior (pipStates, location, etc) should be compared to the old logic, but internal variables (unseenCard, matchingCards, matchingCardsArray) shouldn't. Hence, I suggest that I will change the script to match the new logic, and will verify that the only changes (in the current tests) are in the new variables.

client/test/testMetadata.ts
```diff
@@ -15,6 +15,7 @@ export default function testMetadata(
     },
     ourPlayerIndex: 0,
     spectating: true,
+    replay: false,
```
robertos, line 18: Why this new metadata? If it's an in-game replay (spectating = false) shouldn't be an issue. If it's a post-game replay, we should be turning on spectating and recalculating the whole game from the beginning (which is a current bug too).

asaelr: I added it, as I added the _replay_ field to _GameMetadata_.
This, in turn, was done to solve #1446 - cardReducer must check if it's a replay to calculate _handsSeeingCardForFirstTime_. (check _spectating_ is not enough)

robertos: I think there's a disconnect here - on a post-game replay (not in-game) you should always be spectating. Maybe a better name for the flag is "playing" and the value is the opposite of what it is now.

asaelr: Well, currently on a post-game replay spectating===false. (That's what creates that bug...)

robertos: This is now fixed

robertos:
Re: matchingCards vs Array - the matchingCards map is only used in 3 places in your code right now: 2 of these are in the isPotentiallyPlayable function, and one is on checkNoteImpossibility. I really don't think it's necessary :) We can totally use includes() in these cases, which is an O(n) with n < deckSize.
Better to call matchingCards the array of matching cards (currently matchingCardsArray). A possible better name for it is possibleCardsFromClues, and unseenCards would become possibleCardsFromObservation.

Re: pipStates - i'd love to remove pipstates from the state completely, because it's trivial and fast to calculate them from the real information (the 2 arrays above). That will also make the whole clueMemory obsolete.

Re: tests - i agree that it's probably OK to match only the visible data (pips), but i'm only OK with that if we have a bigger test suite. As I said it's very easy to break this in subtle ways.

asaelr:
> Re: matchingCards vs Array - the matchingCards map is only used in 3 places in your code right now: 2 of these are in the isPotentiallyPlayable function, and one is on checkNoteImpossibility. I really don't think it's necessary :) We can totally use includes() in these cases, which is an O(n) with n < deckSize.
OK. I will get rid of that array. But instead of _for (playableCard) checkIfMatch()_ (which calls includes() about suits.length times) we probably should do _for (matchingCards) checkIfPlayable()_.
> Better to call matchingCards the array of matching cards (currently matchingCardsArray). A possible better name for it is possibleCardsFromClues, and unseenCards would become possibleCardsFromObservation.
I will rename them
> 
> Re: pipStates - i'd love to remove pipstates from the state completely, because it's trivial and fast to calculate them from the real information (the 2 arrays above). That will also make the whole clueMemory obsolete.
sounds correct. we also use _possibilities_  very few times, and we only need _positiveClues_ for rank clues (to highlight them on pink cards), so I will get rid of clueMemory.
> 
> Re: tests - i agree that it's probably OK to match only the visible data (pips), but i'm only OK with that if we have a bigger test suite. As I said it's very easy to break this in subtle ways.
I also verified that _possibleCards === matchingCards ? unseenCards : 0_, but I see your point. However, I really don't want to copy the clues data from the current hanabi.live, as this is exactly the bugged think I'm trying to fix...
Probably (I'm not sure) the current code has the correct _possibleCards_ (as in the definition above), so we can verify that new tests match that.

client/test_data/up_or_down_turn5.json
asaelr: I implemented those changes.
I suggest again, that maybe we do want the state to have something similar to pipStates. Right now we recalculate the pip states of every card when a card is drawn (except of cards in the same hand). (And this is a ui calculation, so we can't put it in a reducer)

client/src/game/reducers/cardPossibilitiesReducer.ts
```diff
@@ -21,219 +14,71 @@ const cardPossibilitiesReducer = (
...
-    }
-
-    // Remove all the possibilities we found
-    const possibilitiesToRemove = (impossibleCards.map((c) => ({
-      suitIndex: c.suitIndex!,
-      rank: c.rank!,
-      all: true,
-    })));
-    possibleCards = removePossibilities(possibleCards, possibilitiesToRemove);
+  const clueTouch = variant.touchedCards.get(clue.value)!;
```
robertos, line 26: as explained above, use cluesRules.touchesCard instead

client/src/game/ui/HanabiCard.ts
```diff
@@ -858,25 +863,26 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   }
 
   // Update all UI pips to their state
-  updatePips(clueType: ClueType | null = null) {
+  updatePips() {
+    enum PipState {Hidden, Eliminated, Visible} // the order is important here
```
robertos, line 867: the order shouldn't matter on an enum :( If you're hiding numbers on enums we can fix it

client/src/game/ui/notes.ts
```diff
@@ -195,19 +195,13 @@ export const checkNoteImpossibility = (variant: Variant, cardState: CardState, n
...
-      if (count > 0) {
-        suitPossible = true;
-        break;
-      }
-    }
-    if (!suitPossible && cardState.location === globals.playerUs) {
+    if (
+      cardState.possibleCardsFromClues.some(
+        ([suitIndex, rank]) => suitIndex === note.suitIndex
+          && cardState.possibleCardsFromObservation[suitIndex][rank] > 0,
```
robertos, line 201: this should probably become a function in rules/card.ts, possible name: "cardRules.canPossiblyBe(cardState, suitIndex, rank)

client/src/game/ui/HanabiCard.ts
asaelr: I use Math.max on them. I will write numbers explicitly.

client/src/game/ui/notes.ts
```diff
@@ -195,7 +195,15 @@ export const cardIdentityFromNote = (
 // Validate that the note does not contain an impossibility
 export const checkNoteImpossibility = (variant: Variant, cardState: CardState, note: CardNote) => {
   // Validate that the note does not contain an impossibility
-  if (!(cardState.location === globals.playerUs)
+  if (cardState.rank === STACK_BASE_RANK
```
Zamiell, line 198: each condition needs to be on a separate line

Zamiell:
```
   if (
      playStacksRules.nextRanks(playStacks[suitIndex], playStackDirections[suitIndex], deck)
        .includes(rank)
    ) {
      return true;
    }
```
-->
```
  const nextRanksArray = playStacksRules.nextRanks(playStacks[suitIndex], playStackDirections[suitIndex], deck);
   if (nextRanksArray.includes(rank)) {
      return true;
    }
```

Zamiell:
very nice work

PR #1460, opened 2020-07-14
Replay the game as spectator when it ends

robertos:


client/src/game/ui/gameCommands.ts
```diff
@@ -289,7 +289,7 @@ commands.set('init', (data: InitData) => {
   const metadata: GameMetadata = {
     options: data.options,
     ourPlayerIndex: data.seat,
-    spectating: data.spectating,
+    spectating: data.spectating || data.replay || data.sharedReplay,
```
Zamiell, line 292: should just be data.spectating || data.replay

robertos: works for me

PR #1461, opened 2020-07-14
Change rank pips to look like numbers

robertos:


client/src/game/ui/HanabiCardInit.ts
```diff
@@ -4,7 +4,9 @@ import Konva from 'konva';
 import * as KonvaContext from 'konva/types/Context';
 import { RectConfig } from 'konva/types/shapes/Rect';
 import * as KonvaUtil from 'konva/types/Util';
-import { CARD_H, CARD_W } from '../../constants';
+import {
+  CARD_H, CARD_W, CLUED_COLOR, CHOP_MOVE_COLOR, FINESSE_COLOR,
```
Zamiell, line 8: when more than 3, must go one on each line

client/src/game/ui/HanabiCardInit.ts
```diff
@@ -255,11 +257,16 @@ export function pips(variant: Variant) {
...
+      fontStyle: 'bold',
+      fontSize: 63,
+      align: 'center',
+      text: rank.toString(),
       width: Math.floor(CARD_H * 0.1),
       height: Math.floor(CARD_H * 0.1),
       fill: 'black',
       stroke: 'black',
-      strokeWidth: 4,
+      strokeWidth: 2,
```
padiwik, line 269: i think you need to scale this against card size

robertos: possible

padiwik: oh seems it's not necessary, since the card gets drawn as an image. For example, the finesse borders scale properly even though their width is constant

PR #1473, opened 2020-07-15
Add shadow effects, remove textured backgrounds + other bugs

robertos:
Bugs related to: ClueEntry not working, listening issues, make FitText faster with a binary search
Perf suffers from the shadows, I'll investigate before moving forward with this. But it does look good :)


Zamiell:
thank you!

Issue #1476, opened 2020-07-15
card-reducer: Suit pips get removed when there's too many negative rank clues


padiwik:
Game 150856, turn 22. The b4 and p5 in dapisani's hand look fine. Go forward a turn, when 5 is clued to y5, and they are missing the suit pips!

Zamiell:
nice bug lol

Zamiell:
confirmed fixed, thanks rob

Issue #1474, opened 2020-07-15
card-reducer: If you mouse over a clued card and then put it back, it will no longer be raised


padiwik:
at least in a hypo, i didn't try in game

Zamiell:
in a game, it immediately stops the raising as soon as you mouse over the card

Zamiell:
this is a regression from 604ba77eac0ab4553e26a85ced8fc34d260d3284

Zamiell:
the game lags out when beginning to drag a card, rob is this because of the shadows?

padiwik:
> in a game, it immediately stops the raising as soon as you mouse over the card

yeah exactly, that's a better way to put it

Zamiell:
confirmed fixed, thanks rob

PR #1477, opened 2020-07-16
Fix bug on terminate + bug when a card is clued and gets mouse over

### LINKED ISSUE #1475: card-reducer branch: rob should fix case 'cardidentities' in stateReducer.ts

robertos:
Fix #1475  #1474


robertos:
+ Fix #1476

Zamiell:
nice!

Issue #1478, opened 2020-07-16
card-reducer - bug, cards get played at their offset


Zamiell:
![image](https://user-images.githubusercontent.com/5511220/87676832-1492ec80-c747-11ea-86e1-441abc63067c.png)

Zamiell:
seems to be fixed

PR #1483, opened 2020-07-17
Fix animations with shadow offset

robertos:


robertos:
Fix #1478

Zamiell:
nice, thanks

PR #1500, opened 2020-07-19
Fix Throw it in a Hole

### LINKED ISSUE #1385: card-reducer branch: make throw it in a hole work

robertos:
Fixes #1385


Zamiell:
well that looked a lot more complicated than just the animation

Issue #1489, opened 2020-07-17
card-reducer branch - bottom-deck-blind-plays don't animate


Zamiell:
repro:

```
[...]

PR #1507, opened 2020-07-20
Fix bottom deck blind plays on replay / if you're not the player who played it

robertos:


robertos:
Actually fix #1489

Zamiell:
thanks

PR #1515, opened 2020-07-20
Set RankPips to the suit color if available

### LINKED ISSUE #1505: card-reducer branch - number pips are not great with some suits

robertos:
Fix #1505


Zamiell:
thanks

PR #1537, opened 2020-07-27
Fix text updates while action log is open

robertos:
Repro: during a replay, open the full action log or a player log, press arrow left/right
Expected: text updated
Actual: text disappears


Zamiell:
thanks

Issue #1544, opened 2020-07-28
Missing-score page errors out when filtering by number of players


OnceUponAQuine:
![image](https://user-images.githubusercontent.com/65525468/88612294-c4b5fe80-d03f-11ea-964a-077b9acf79b7.png)

PR #1545, opened 2020-07-28
Fix issue 1544

OnceUponAQuine:
https://github.com/Zamiell/hanabi-live/issues/1544

Also made the input field for score a number input instead of a text input because I noticed it. Feel free to cherry-pick `1a1c0dc` if you just want the bugfix.


Zamiell:
thanks

PR #1550, opened 2020-07-28
Ensure immer auto-freezes deeply recursive structures + proper sourcemaps for libs

robertos:
200x performance increase for reducers! pogchamp


Zamiell:
nice thanks

PR #1572, opened 2020-07-30
Fix border and offset bugs + faded X opacity

### LINKED ISSUE #1566: Sometimes finesse/clued/cm borders disappear after/during replays
Don't have a good repro yet

### LINKED ISSUE #1567: Sometimes cards that were clued end up offset (in Y) on playstack
No good repro yet. Seems related to quickly clicking on cards after dragging?

robertos:
Fix #1566 and likely fix #1567


Zamiell:
nice thanks

PR #1604, opened 2020-08-09
Fix raising bug and Konva error

### LINKED ISSUE #1598: If you clue a card, then play it, then rewind (hold left arrow), card can stay raised on turn before it was clued
It's related to the "unclue" happening before the animation where it goes back to the hand finishes.
E.g.
Clue a card on turn 2
Play it on turn 5
Hold left
If the animation of the card moving back from the playstack is still happening when the replay goes from turn 2 to 1, then the card stays raised.

Also repros on post game / JSON replays.

### LINKED ISSUE #1602: konva error when bar dragging to turn 1

robertos:
Fix #1602 and fix #1598


Zamiell:
thank you!

Issue #1469, opened 2020-07-14
Proposal: Sudoku Inferences


Smandon1:
Empathy makes inferences on your cards, such as removing pips based on seeing other cards out on the table (I think this might be #427).  I think it's neat and wanna see more :P.  I propose an additional layer of inferences (maybe some of this exists already?), similar to solving a sudoku:

If N slots can only be a set of N cards then remove those cards from the possible cards for the other slots.  For example (no variant):
a) N=1: There are three unplayed fives: red, yellow, and blue.  Alice has two fives in slot 1 and 5.  The five in slot 5 has a negative yellow clue.  When Bob draws a red five, that makes Alices fives cross out red from both of her fives.  The only possibility for Alice's five in slot 5 is blue (N=1).  As a result, this should remove blue five from the possible cards for Alice's slot 1 five (since there are no other copies of blue fives).
b) N=2: (replay [here](https://hanabi.live/replay/211426), Smandon's hand on turn 48) All blue cards are seen/known by Alice except a single blue 2 and a single blue 5.  Alice has an existing blue clue on her slot 2 and 3 cards.  One of them has to be a blue 2 and the other has to be a blue 5.  Two slots can only be two cards, therefore the other slots cannot be those two cards and can have them removed as possible cards from Alice's other slots.  This would help slot 1 if the card was drawn after the blue clue.  In the replay, that should mean slot 1 empathy can show green 5.  This is akin to the [naked pair](https://www.learn-sudoku.com/naked-pairs.html) in sudoku.

This would be done iteratively until no new inferences can be made.  Since the max hand size is like 6 cards, I would guess that it shouldn't really be much of a performance hit, if any.  I'd also assume checking/iterating over the N=1 cases first before incrementing N would improve performance if that's a concern.

Zamiell:
i'm pretty sure that this is a duplicate of #1459

robertos:
No, it's not the same thing. This is talking about making the logic recursive. That one is to make it retroactive

asaelr:
It's more than recursive. (recursion is the case N=1)
I would consider implementing this only for the final round. This is when it's most likely to give a useful information. (Even though other situation are possible - for example if I see r5,b5, and I have three cards with a positive 5 clue, and two of them have a negative green).

Smandon1:
@asaelr 
> I would consider implementing this only for the final round.
Is that due to performance?  I'm hopeful that it should be performant at any point in the game, given the small hand size.

[...]

Issue #10, opened 2018-01-07
ingame chat


Zamiell:

Zamiell:
you click a "chat" button, and it slides over from the right, covering the clue log + the discard pile
and then u click X in the top right hand corner and it slides back to the way it was

Zamiell:
need to make it a floaty text box + history

Zamiell:
floaty done, now i need to make it popup on new msg

Zamiell:
done

PR #1608, opened 2020-08-10
Using Deduction To Reduce Card Possibilities

Smandon1:
This is for issue #1469.  It's just trying out all of the possibilities, as opposed to projecting an inference based on pairs / triplets /etc of cards.

Hopefully performance isn't too bad.  The only place I noticed it was on loading a reply, it takes a little bit longer.  If that becomes an issue we can try to speed up the deduction, put the (faster) inference logic in front of the deduction, or can switch to just using an inference algorithm.


client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
+import * as deckRules from '../rules/deck';
+import { GameAction } from '../types/actions';
+import CardState from '../types/CardState';
+import GameMetadata from '../types/GameMetadata';
+import Variant from '../types/Variant';
+
+const cardDeductionReducer = (
```
robertos, line 7: The parameter order is wrong for this to be a reducer.
newThing = reduce(oldThing, action, other params)
so
newDeck = cardDeductionReducer(oldDeck, action, hands, metadata)
Note metadata already includes the variant name, no need to pass again.

client/src/game/reducers/integration.test.ts
```diff
@@ -130,6 +158,34 @@ describe('integration', () => {
         expect(finalState.stats.potentialCluesLost).toBe(18);
       });
     });
+
+    describe.each([...Array(rainbowOnesAndPinkGame.actions.length).keys()])('on turn %i', (turn) => {
+      test.each([...Array(45).keys()])(
```
robertos, line 163: this is not an integration test - would be better to get it as a unit test with a single turn, with a few carefully chosen cards.
an integration test for this is what we're doing now, which is to compare an expected final state with the actual result we got.

client/test_data/up_or_down_final_cards.json
```diff
@@ -4,71 +4,133 @@
         "location": "discard",
         "suitIndex": 0,
         "rank": 4,
-        "possibleCardsFromObservation": [
```
robertos, line 7: if we keep changing the baseline, this test is a lot less useful. this test is to make sure changes like this one don't break existing logic, so we can't keep changing the baseline every time.

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export default cardDeductionReducer;
+
+const makeDeductions = (
+  hands: ReadonlyArray<readonly number[]>,
+  deck: readonly CardState[],
+  variant: Variant,
+  metadata: GameMetadata,
+) => {
+  const newDeck: CardState[] = deck.concat([]);
```
robertos, line 36: Array.from is cleaner than concat([])

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
...
+  switch (action.type) {
+    case 'cardIdentity':
+    case 'clue':
+    case 'discard':
+    case 'play':
+    case 'draw': {
+      return makeDeductions(hands, deck, variant, metadata);
+    }
+    default: {
+      return deck.concat([]);
```
robertos, line 23: Array.from is cleaner

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
...
+    return true;
+  }
+  cardCountMap[suit][rank] += 1;
+  return false;
+};
+
+const remainingCardsValid = (
+  remainingCards: readonly CardState[],
+  cardCountMap: readonly number[][],
+) => remainingCards.length === 0 || remainingCards[0].possibleCardsFromDeduction.some(
```
robertos, line 107: break this into multiple lines

robertos: Also why is this passing an array of card states around? card states are very big, so this is basically making multiple copies of the deck?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
...
+  (possibility) => possibilityValid(possibility, remainingCards.slice(1), cardCountMap),
```
robertos, line 108: this is a pretty obtuse way to implement recursion, by creating 2 functions that call each other. I think a classic "base case/regular case" recursive function would be much cleaner.

client/src/game/reducers/cardDeductionReducer.ts
robertos: (on every action)

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Otherwise, it's a card in another hand and we know we don't have that copy.
+    if (deckCard.suitIndex === null || deckCard.rank === null || otherHand === hand) {
+      remainingCards.push(o);
+    } else {
+      remainingCardCountMap[deckCard.suitIndex][deckCard.rank] -= 1;
+    }
+  }));
+  hand.forEach((o) => {
+    const card = deck[o];
+    const otherRemainingCards = remainingCards.filter((ro) => ro !== o).map((ro) => deck[ro]);
```
robertos, line 79: what you really want here is map the order with the possibleCardsFromDeduction from the other, not the whole cardstate

robertos (dismissed):

robertos:
If you think "it takes a bit longer" then we definitely need to measure. The whole game processing should take less than 20ms now, so if you can see it then it's a 10x or more impact :/

Smandon1:
Hopefully I'm replying to the "it takes a bit longer" thread.  My computer might be a bit slower than yours (or I could just be measuring the wrong thing).  Here's what I measured, if I was doing it right:
I loaded the replay json for game 211426 ten times with this pull request and ten times with the original code (right before the pull request).  I looked at the chrome performance for the reduceGameActions method:

old code (in ms):
129.89
113.90
114.51
120.95
118.04
111.22
121.13
114.69
120.96
118.50

new code (in ms):
275.32
331.76
279.06
275.89
330.23
321.98
341.26
306.16
336.77
329.69

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: I don't disagree that I only need possibleCardsFromDeduction for this method, so I'll fix that.  I do want to understand the comment about copies better.  I thought javascript was pass-by-value of the reference and therefore wasn't copying the whole CardState, just copying the reference value of the CardState (or in this case, a copy of the reference to the array).

If I add a `remainingCards.pop()` call to the end of the `possibilityValid` method, then tests fail.  If it were copying the entire array, as opposed to copying the reference value, then I would think it would have no impact (since it's at the end of the method)?

```
const possibilityValid = (
  [suit, rank]: readonly [number, number],
  remainingCards: CardState[],
  cardCountMap: readonly number[][],
) => {
  // Avoiding duplicating the map for performance, so trying to undo the mutation as we exit
  cardCountMap[suit][rank] -= 1;
  if (cardCountMap[suit][rank] >= 0 && remainingCardsValid(remainingCards, cardCountMap)) {
    cardCountMap[suit][rank] += 1;
    return true;
  }
  // We're no longer using remainingCards inside this method after the pop call
  remainingCards.pop();
  cardCountMap[suit][rank] += 1;
  return false;
};
```

client/src/game/reducers/integration.test.ts
Smandon1: My intention was for it to be more of a sanity check - making sure the logic doesn't mess up so badly that it somehow removes the correct suit/rank.  I was also thinking about running through the game for each player (so removing knowledge of your own hand from the underlying logic) with the same check.  It's not really making a positive check (saying "it removed this correct suit/rank"), so maybe that's not valuable :/.  I can remove it.

For unit tests: I added some tests under `cardsReducer.test`.  I sort of treated it as though `cardDeductionReducer` was a part of `cardsReducer` when running tests but I can see an argument that it's not really a unit test at that point.  Would it be best to mock the `cardDeductionReducer` in `cardsReducer.test` and just move those tests over to a `cardDeductionReducer.test` (asserting a single turn)?

client/test_data/up_or_down_final_cards.json
Smandon1: I get that, although I'm not sure what I should be doing instead in this case...  I certainly wasn't setting out to lose the check we had before.  I don't think you're suggesting I need to make a change in this particular instance, if I understand correctly.  Just saying "we gotta stop doing this all the time"?

With the full deduction "check if it's impossible to be this card", there's not really a count of each suit/rank anymore.  It's either possible or it's not.  So I don't really have access to / a need for a count.  If we added some inference back in (perhaps for performance) and kept deduction then that'd probably be a third layer, in between `possibleCardsFromClues` and `possibleCardsFromDeduction`, with a map of suit/rank to count.

One option would be to still include the `possibleCardsFromObservation` assertion values from prior to this commit and in code convert it to a suit/rank list to assert against `possibleCardsFromDeduction`, so we can see which counts changed from non-zero to zero but that feels a little clunky to be carrying around counts if the logic isn't differentiating between the non-zero values.

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -18,23 +18,23 @@ const cardDeductionReducer = (
     case 'discard':
     case 'play':
     case 'draw': {
-      return makeDeductions(current(hands), deck, variant, metadata);
+      return makeDeductions(deck, current(createDraft(hands)), metadata);
```
robertos, line 21: don't use createDraft here. Are you passing hands that are not draft to it? Maybe a better solution is to use current() on the caller (cardsReducer)

Smandon1: I don't really understand immer, so that doesn't help :).  My tests in `cardsReducer.test.ts` were passing hands that weren't draft.  I tried to do `createDraft` under `initialGameState` but that didn't work.  I think I was maybe missing tests that overwrote hands.  That said, maybe the tests shouldn't be using `createDraft` either and should be doing something different to get the `current` call to work.

robertos: The right solution is to ensure the call to current happens at cardsReducer (the caller to this). Then your tests don't have to worry. You should never do createDraft.

Smandon1: The tests for cardsReducer still fail (complaining that it's not a draft).  If I move it up further to gameStateReducer then it still complains in the integration test.  I can probably just mock the `current` method in my tests and that wouldn't be a very big change.

I should note that I'm testing `cardDeductionReducer` via `cardsReducer.test` (kind of treating them as one unit).  If you'd like, I can mock `cardDeductionReducer` in `cardsReducer.test` and make a separate `cardDeductionReducer.test`.  It still wouldn't solve the `current` issue though because I believe the remaining tests would fail in `cardsReducer.test` (or `integration.test`).

Is `current` supposed to speed the code up?  The [docs](https://immerjs.github.io/immer/docs/current) for `current` mainly seem to say that it's useful to call that method for debugging?  It's taking a (mutable) draft and creating another (mutable) copy?  You probably don't need to delve too deep into the details here... I just don't get the impression that `current` is supposed to improve performance when reading the docs.

robertos: I think there's some misuse of drafting. The perf benefit is that reading from drafts is REALLY slow (they are proxies). Current converts a draft into the real object.

robertos: You can solve it definitely by having whatever reducer that changes hands into a draft (likely cardsReducer) call current when it's done.

Smandon1: I think `gameStateReducer` creates a draft of `GameState`, which `cardsReducer` has passed in as a parameter.  I think I was putting `current` around the wrong thing when I previously tried to put it in `gameStateReducer`.  I'm not sure if we want to copy the whole `GameState` out of draft when calling `cardsReducer` or not.  I put a `current` call in `cardsReducer` when calling `game.hands` and then mocked out the `current` call in `cardsReducer.test` and that seems to work.

client/src/game/reducers/cardsReducer.test.ts
```diff
@@ -24,6 +24,9 @@ const fourthCard = initialCardState(3, variant);
 const fifthCard = initialCardState(4, variant);
 
 jest.spyOn(deckRules, 'isInitialDealFinished').mockReturnValue(true);
+jest.mock('immer', () => ({
+  current: jest.fn().mockImplementation((a: ReadonlyArray<readonly number[]>) => a),
```
robertos, line 28: mocking immer sounds very, very wrong. I'll look and see where is the mistake

Smandon1:
I'm still kinda new to PR's.  Do I need to do anything more here on my end ("resolve conversation", etc)?  I think I like where the code is at.  We discussed changing the integration test a bit for the long term (using a 'view' dto).  Not sure if that belongs in this PR or a separate one.  I don't know how quick it would be to get the integration test changed over.

robertos:
@Smandon1 you don't have to do anything, but I want to test this more (particularly perf) and see if I can make the change I told you before merging.

Smandon1:
ok cool, no worries

Zamiell:
rob did you get a chance to finish reviewing this?

Zamiell:
earth to rob

Zamiell:
@Smandon1 rob is currently moving so he doesn't have a lot of time to devote to hanabi live.
i have no idea what your pr is doing so lets wait for him to come back

Smandon1:
Sounds good, works for me

Smandon1:
@Zamiell I've got a bunch of vacation time at the last half of December and I could get back at this, bringing it back up to the main branch.  Do you know if rob will be coming back anytime soon / if I should work with someone else on this?

Zamiell:
i dunno, can you DM him on discord?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,144 @@
...
+  const newDeck: CardState[] = Array.from(deck);
+  const variant = getVariant(metadata.options.variantName);
+  const cardCountMap = getCardCountMap(variant);
+
+  // if our variant shows played/discarded cards, then they're known by everyone
+  newDeck.filter((card) => !hands.some((hand) => hand.includes(card.order)))
+    .filter((card) => card.suitIndex !== null && card.rank !== null)
+    .forEach((card) => {
+      cardCountMap[card.suitIndex!][card.rank!] -= 1;
+    });
```
timotree3, line 44: This can be simplified to use `card.location`. Also, it's expensive to use `filter` when you're just going to iterate over it's result.
```suggestion
  // if our variant shows played/discarded cards, then they're known by everyone
  for (const card of newDeck) {
    if ((card.location === 'discard' || card.location === 'playStack') && card.SuitIndex !== null && card.rank !== null) {
      cardCountMap[card.suitIndex][card.rank] -= 1;
    }
  }
```

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,144 @@
...
+    possibleCardMap[suitIndex] = [];
+    const suit = variant.suits[suitIndex];
+    possibleRanks.forEach((rank) => {
+      possibleCardMap[suitIndex][rank] = deckRules.numCopiesOfCard(
+        suit,
+        rank,
+        variant,
+      );
+    });
+  });
```
timotree3, line 138: Can this be simplified a bit?
```suggestion
  const possibleCardMap: number[][] = [];

  variant.suits.forEach((suit, suitIndex) => {
    possibleCardMap[suitIndex] = [];
    variant.ranks.forEach((rank) => {
      possibleCardMap[suitIndex][rank] = deckRules.numCopiesOfCard(
        suit,
        rank,
        variant,
      );
    });
  });
```

timotree3 (commented):
We can probably merge this without Rob's finishing touches. I left a couple of suggested changes. After addressing those and fixing all the conflicts with recent changes to master, I think we can merge this PR.

A couple of thoughts for future improvements: 

- Right now I don't think the logic takes into account detrimental characters. For example, when looking at the empathy for a Slow-Witted player, if another player had a yellow 5 in slot 1, it would show a yellow 5 possibility as crossed in their hand even though they can't see that card.
- It would be nice to distinguish which deductions are globally known (e.g. from cards in the play stacks or fully identified cards in hands) and instead of crossing out those pips, completely hide them.

Zamiell:
> It would be nice to distinguish which deductions are globally known (e.g. from cards in the play stacks or fully identified cards in hands) and instead of crossing out those pips, completely hide them.

I dont want to make pips invisible unless they are from clues, so that there is a clear delineation between stuff happening from clues and stuff happening from [other]

timotree3:
> > It would be nice to distinguish which deductions are globally known (e.g. from cards in the play stacks or fully identified cards in hands) and instead of crossing out those pips, completely hide them.
> 
> I dont want to make pips invisible unless they are from clues, so that there is a clear delineation between stuff happening from clues and stuff happening from [other]

It seems like what matters is a clear delineation between globally known and known using local information. If something is globally known, why does it matter whether from clues or something else?

Zamiell:
because historically the only way that pips have appeared or disappeared is through clues and nothing else.

regardless of historical precedent, a new player may be confused as to why a particular pip is invisible:

"Oh, that's interesting, the green pip on my card is gone. That means that this card cannot be green. But I have not received a negative green clue. Oh well, it must be a bug in the interface - that's annoying."

in this hypothetical, it is near the end of the game, and Alice sees every green card in the game, which is causing the behavior in question.
but figuring out that you see very green card in the game is non-trivial, for both beginner players and intermediate players alike

timotree3:
Another bug is that in Throw-it-in-a-Hole, while the local player may know the suit index and rank of a card that has been played, because they saw it in another player's hand before it was played, not every player necessarily does, meaning that removing those possibilities from the `cardCountMap` will eliminate more possibilities from empathy then the other player actually sees eliminated.

Zamiell:
should probably just disable this feature if:
1) in a throw it in a hole variant
2) if detrimental characters is turned on

Zamiell:
github reports that branch has conflicts that must be resolved, smandon im assuming thats something you can do?

Smandon1:
Currently, this feature doesn't retain the cardCountMap between turns.  It just has what the possibilities are, without a count.  I think I can get throw it in a hole and detrimental characters (Slow-Witted) to work with the new feature, so that way we don't have to keep code for the old way and the new way.  I'll need to look through the other detrimental characters, looks like there's some other characters that have vision problems (Blind Spot and Oblivious), but I should be able to get those to work too.

Smandon1:
@Zamiell  - Yeah, I'll resolve the conflicts.  This branch is several months behind, and I haven't looked much at the new code / conflicts yet, but I should be able to get it resolved within the next couple of days or so.

client/src/game/reducers/cardDeductionReducer.ts
robertos: There's a very specific reason why it's the way it is now - if you don't forEach it will skip empty fields, and this has to be aligned with 1-5 and 7 (skipping 0 and 6). Try it out and you'll see :)

timotree3: Oh wow good point

Smandon1: Thanks rob, I'll switch it back.

Smandon1:
I think the existing code is probably a bit buggy for empathy on Slow-Witted players.  When a Slow-Witted player draws a card, it will remove all cards in other hands (even if they're in slot 1) from possibleCardsFromObservation.  If another player draws a card, it will remove that possibility from possibleCardsFromObservation on the Slow-Witted players hand, even though the Slow-Witted player wouldn't see that newly drawn card.  Both of those examples assume you are not the Slow-Witted player or the one drawing the card.  Hopefully this new change should fix that as well.

Today we use possibleCardsFromObservation to fuel empathy and to figure out what's in our own hand.  That generally works well because possibleCardsFromObservation applies to the player holding the card.  If it's held by another player then it represents empathy, and if it's held by us then it's just what we know about it (I hesitate to call that self-empathy... not sure what the proper term should be).

This can become a problem when you don't know what a card is in another persons hand.  Today, I think we have it represent empathy for the card holder.  For example, if we're Slow-Witted, playing 3-suit, and see another player with three 5's in slot 2/3/4 then we will not cross out 5 on slot 1 for that player (because that other player wouldn't know they have three 5's).

I can leave that directive the same (have it always be empathy for the card holder).  We could change it to represent our own knowledge of the card (crossing out 5 in the previous example).  I'm not sure if that would get confusing for the Slow-Witted player or not.  I should be able to do either approach relatively easily with this change.

Long term, we potentially could show both.  Default to show our own knowledge of the card but then if that hand is clicked then switch to showing empathy.  I'm not sure what would happen if a player populated a note, "revealing" the card and covering up our knowledge of the card.  That can only happen in our own hand today, and clicking on the card shows our own knowledge.

timotree3:
Seems like it should show what you know about a card by default, and what the holder knows while in empathy mode

Smandon1:
I think I've merged the latest changes into this branch but there's still some TODOs:
1) Try to look over my new code to match the new conventions I noticed (double quote usage instead of single quote, changing functions from being defined as a const variable to instead just be a function).  Let me know if there's other things I should watch out for that you can remember.  Looks like lint should help with that, for some reason it didn't show me these errors locally but I might need to restart vscode or something.
2) I need to split possibleCardsFromDeduction into two, one for empathy of the card holder (maybe possibleCardsForEmpathy) and another for our best knowledge about the card (maybe just possibleCards).  Then use that to allow for different pips to be shown on an unknown card in another persons hand, depending on whether we're in empathy mode (as discussed in the last comment).

Zamiell:
> double quote usage instead of single quote

we use prettier now. so all you have to do is hit "ctrl + s" to save the file and it autoformats single quotes to double quotes, etc

Zamiell:
as you can see by clicking on "Details" below, there are 1000 prettier errors, so it looks like you aren't using prettier.
i've included instructions for getting it setup in INSTALL.md

Zamiell:
all you should have to do is npm install and restart vscode probably

Smandon1:
Yeah, I think the `npm install` and restarting vscode fixed it.  Thanks!

Zamiell:
ya, it passes ci now

client/src/game/reducers/integration.test.ts
Smandon1: I believe these "on turn %i" tests have been removed.

Smandon1:
Both of my prior TODOs are now done.  I've also added a performance enhancement, to not recalculate a card if the other cards it's using for that calculation have not changed.  When loading a replay, it's currently about 80ms slower to load than prod on my machine (to calculate all turns, combined).  When playing a game, that difference should be a lot less since we're only calculating one turn per action.

I think I'm ready for this to be merged, assuming the code and everything looks okay to you guys.

Zamiell:
timo do you want to take a look + make sure all of rob's feedback was addressed?

timotree3:
Sure I'll take another look right now. Thanks again Smandon :)

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+    if (playerIndex === ourPlayerIndex) {
+      possibleCards = possibleCards.filter((possibility) =>
+        possibilityValid(possibility, possibilities, 0, cardCountMap),
+      );
+      if (playerIndex === card.location) {
+        // If the card is in our own hand then we also need to update empathy
+        if (card.revealedToPlayer[playerIndex]) {
+          // If it's revealed to us then we can safely just use possibleCards and don't
+          // need to do a separate calculation run for possibleCardsForEmpathy.
+          possibleCardsForEmpathy = possibleCards;
```
timotree3, line 97: When would a card in our hand ever be revealed to us?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+    )
+  ) {
+    cardCountMap[suit][rank] += 1;
+    return true;
+  }
+  cardCountMap[suit][rank] += 1;
+  return false;
+}
+
+let cachedVariantName: string | null = null;
```
timotree3, line 316: Can we cache the variant ID instead?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+  excludeCardOrder: number,
+  deck: readonly CardState[],
+  playerIndex: number,
+) {
+  const unknownCards = [];
+  for (const card of deck) {
+    if (card.order === excludeCardOrder) {
+      continue;
+    }
+    if (card.revealedToPlayer[playerIndex] || card.hasClueApplied) {
```
timotree3, line 235: Does this mean that cards in a player's own hand don't get brought in for deduction?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+    playerIndex,
+  );
+  const possibilities: Array<ReadonlyArray<readonly [number, number]>> = [];
+  for (const card of unknownCards) {
+    possibilities.push(
+      getCardPossibilitiesForPlayer(card, playerIndex, ourPlayerIndex),
+    );
+  }
+  // start with the more stable possibilities
+  possibilities.sort((a, b) => a.length - b.length);
```
timotree3, line 221: Why does this matter?

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+            (possibility) =>
+              possibilityValid(possibility, possibilities, 0, cardCountMap),
+          );
+        }
+      }
+    } else if (playerIndex === card.location) {
+      possibleCardsForEmpathy = possibleCardsForEmpathy.filter((possibility) =>
+        possibilityValid(possibility, possibilities, 0, cardCountMap),
+      );
+    }
```
timotree3, line 112: I find the logic in these ifs a bit hard to read/understand. Maybe it can be simplified to something like this?
```suggestion
    if (playerIndex === ourPlayerIndex) {
      possibleCards = possibleCards.filter((possibility) =>
        possibilityValid(possibility, possibilities, 0, cardCountMap),
      );
    }
    if (playerIndex === card.location) {
      possibleCardsForEmpathy = possibleCardsForEmpathy.filter((possibility) =>
        possibilityValid(possibility, possibilities, 0, cardCountMap),
      );
    }
```

timotree3 (changes requested):
This new approach seems robust with respect to Detrimental Characters and Throw-it-in-a-Hole. Nice work :)

Performance: I think we're running this deduction code far more often than we need to. We're recomputing it upon every action, without being able to reuse any of the past results. Ideally we would only do the deduction on states that we actually intend to display. For example, if someone enters a replay, it just calculates the deductions for the turn that they're on.

What do you think about the possibility of moving this deduction code into the UI so that it runs on-demand? Maybe we would store the results in a global or something instead of in the Redux state.

client/test_data/up_or_down_final_cards.json
timotree3: I think we can safely resolve this. It's unfortunate that the test isn't really helping us for this change but I think that's natural given that we changed the data structure and the test cares about the structure of the output.

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+    // If this card is in the players hand, then use our best (empathy) guess.
+    return card.possibleCardsForEmpathy;
+  }
+  if (playerIndex === ourPlayerIndex || card.revealedToPlayer[playerIndex]) {
+    // This is revealed to the player or we are the requested player => just use our best knowledge.
+    return card.possibleCards;
+  }
+  // This is an unrevealed card not in the players hand but not revealed to them.
+  // That can happen with something like a detrimental character (such as 'Slow-Witted')
+  // or throw it in the hole.  We can't use our best (empathy) guess, because it might be in our own
```
Zamiell, line 197: 1) its throw it in a hole
2) capitalize it appropriately
3) one space between periods

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+  card: CardState,
+  playerIndex: number,
+  ourPlayerIndex: number,
+): ReadonlyArray<readonly [number, number]> {
+  if (card.location === playerIndex) {
+    // If this card is in the players hand, then use our best (empathy) guess.
+    return card.possibleCardsForEmpathy;
+  }
+  if (playerIndex === ourPlayerIndex || card.revealedToPlayer[playerIndex]) {
+    // This is revealed to the player or we are the requested player => just use our best knowledge.
```
Zamiell, line 192: don't use periods for single sentences

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+    deck,
+    playerIndex,
+  );
+  const possibilities: Array<ReadonlyArray<readonly [number, number]>> = [];
+  for (const card of unknownCards) {
+    possibilities.push(
+      getCardPossibilitiesForPlayer(card, playerIndex, ourPlayerIndex),
+    );
+  }
+  // start with the more stable possibilities
```
Zamiell, line 220: capitalize the first word of a comment

client/src/game/reducers/cardDeductionReducer.ts
timotree3: I think you're right Smandon. Passing arrays around should be as performant as passing integers.

client/src/game/reducers/cardDeductionReducer.ts
Zamiell: should be else if

timotree3: My suggestion should be else if? No. We still need to update `possibleCardsForEmpathy` on our own cards

client/src/game/reducers/cardDeductionReducer.ts
Zamiell: note that it depends whether you are passing a vanilla array or a redux draft array.
if you are using redux anything arrays, then i think its worse

client/src/game/reducers/cardDeductionReducer.ts
Zamiell: o

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: It won't, that's a bad if check. I initially had it just if it's a card in our own hand then make possibleCardsForEmpathy = possibleCards. That caused replays to have bad empathy in our own hand because possibleCards is just one thing (the real value of the card). So I attempted to fix it by saying "don't do that if the card isn't revealed to us" but that just caused everything to just go down the path of calculating empathy separately.

I've updated it to your simplified logic. Possibly at some point in the future there's a way to split those. It might not be worth the complexity though, since replays are where we take the biggest performance hit and it needs to be run separately for that.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Done

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Done. Some details are in a comment above about why it was (erroneously) like that previously.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Done. I'm apparently behind the times with the "one space between periods" thing. Hopefully that habit won't be too tough to break.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Done

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Done

client/src/game/reducers/cardDeductionReducer.ts
Zamiell: https://www.instructionalsolutions.com/blog/one-space-vs-two-after-period#:~:text=Nearly%20all%20style%20guides%20agree%20that%20one%20space%20is%20correct.&text=The%20Publication%20Manual%20of%20the,space%20in%20its%202019%20update.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: This is for performance. It seemed to have a measurable difference.

The possibilityValid method will short-circuit if it finds a branch that's impossible or if it finds a possibility that's valid. Here's an attempt at an example:

possibilities = [[red 5 or yellow 5],[green or blue],[green or blue],[green or blue],[red 5]]

possibilityValid would initially start with the first card being red 5. It would then check about 1000 combinations of the next three cards before finding each one is impossible at the very end of each combination. If we reorder that to:

possibilities = [[red 5],[red 5 or yellow 5],[green or blue],[green or blue],[green or blue]]

Then when it attempts to resolve [red 5 or yellow 5] to red 5, it will realize that's impossible and short-circuit that branch (not checking the next 1000 combinations of the next 3 cards). It would switch to checking if the combination would work when [red 5 or yellow 5] resolves to yellow 5 (by finding a combination of the next 3 cards that fit).

So it should get to a fitting combination quicker or find that there is no fitting combination quicker.

Smandon1: This applies to more than just cards that have one possibility (such as red 5 in the previous example).

Zamiell: i mean all of what you wrote should be sticked into a comment above the code in question then

Smandon1: Yeah, that's fair. This assumes that I've explained it well enough here.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: There's two sets of cards. I'm bad at naming, but in the code I called them `cardsToCalculate` and `unknownCards`. The `cardsToCalculate` are the cards that we are trying solve the latest `possibleCards` and `possibleCardsForEmpathy` values for. The `unknownCards` are the cards that we're using to disprove as many possibilities (`possibleCards` and/or `possibleCardsForEmpathy`) of each card in `cardsToCalculate`.

A card in the player's own hand might not be added to `unknownCards` because it has no positive or negative clues on it. It would unnecessarily expand the search tree of the `possibilityValid` method because it provides no constraint against the possibility we're checking on the card we're calculating (the unrevealed/unclued card can practically be anything... it would never disprove something). So this improves performance by trimming that search tree.

The `card.order === excludeCardOrder` check is making sure the card we're currently calculating for is not one of the `unknownCards`. Otherwise, we'd be double counting it in the `possibilityValid` method.

If that makes sense, I can try to sprinkle some comments around the code to this effect (or just take those two paragraphs and put it above the highlighted code from the `generateUnknownCards` method).

client/src/game/reducers/cardDeductionReducer.ts
timotree3: Oh this makes sense especially now that I understand the comparison function. I thought `a.length - b.length` was trying to compare how the number of possibilities had changed since last deduction but it's just comparing two elements of the same list

client/src/game/reducers/cardDeductionReducer.ts
timotree3: Makes sense. Does that mean that a card that is revealed is considered an unknown card rather than just subtracted from the card count map beforehand?

timotree3:
@Smandon1 did you see the top-level review comment?

> Performance: I think we're running this deduction code far more often than we need to. We're recomputing it upon every action, without being able to reuse any of the past results. Ideally we would only do the deduction on states that we actually intend to display. For example, if someone enters a replay, it just calculates the deductions for the turn that they're on.
> 
> What do you think about the possibility of moving this deduction code into the UI so that it runs on-demand? Maybe we would store the results in a global or something instead of in the Redux state.

Smandon1:
Yeah, I haven't got to all the comments yet. There is some reuse from action-to-action, although it could be faster running on-demand since we'd just be running for one action (even though we wouldn't be able to reuse data from the previous action). I'll write up a bit more on the reuse later, if it matters. I'm not opposed to generating possibleCardsFromEmpathy on-demand when empathy is turned on.  possibleCards should probably be precalculated / run within the redux stuff, so that way you can quickly page through turns/actions quickly and see the cards fly around. I'd need to look more at that HanabiCard class and how it's created. It currently doesn't have reference to the deck but likely would need it? I don't think the performance is that bad though within the redux stuff.

> @Smandon1 did you see the top-level review comment?
> 
> > Performance: I think we're running this deduction code far more often than we need to. We're recomputing it upon every action, without being able to reuse any of the past results. Ideally we would only do the deduction on states that we actually intend to display. For example, if someone enters a replay, it just calculates the deductions for the turn that they're on.
> > What do you think about the possibility of moving this deduction code into the UI so that it runs on-demand? Maybe we would store the results in a global or something instead of in the Redux state.

Smandon1:
In regards to "What do you think about the possibility of moving this deduction code into the UI so that it runs on-demand? Maybe we would store the results in a global or something instead of in the Redux state."

I think my current preference would be to have it calculated up-front like it is today, although I don't know if I'll be convincing enough:
1) I've sped up the code more with the addition of the 'possibilitiesToValidate' list / logic.  The current code is now about 35ms slower than prod.  I don't think 35 ms is a very big or noticeable jump to end users.  That difference is probably higher than in the wild, since running the performance tab in chrome probably slows things down (expands the difference).  When this code first started, I think it was about 300ms slower.
2) I'm a little worried about rewriting tests (the time it would take) or losing the sanity check of the tests if we ditch them.  I think the integration test values are for every card, whatever the values were the last time they were in someone's hand.

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,343 @@
...
+
+function calculatePlayerPossibilities(
+  playerIndex: number,
+  ourPlayerIndex: number,
+  hands: ReadonlyArray<readonly number[]>,
+  deck: CardState[],
+  oldDeck: readonly CardState[],
+  cardCountMap: readonly number[][],
+) {
+  const cardsToCalculate = getCardsToCalculate(
```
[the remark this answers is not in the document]
Smandon1: I got rid of the cardsToCalculate and unknownCards list in favor of iterating directly.  The generateDeckPossibilities method (previously generatePossibilitiesForUnknownCards) still creates a list for deckPossibilities, since that's needed to sort the values within before iteration.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: Since I got rid of the `unknownCards` list, in favor if directly iterating, I changed the higher level name/list to `deckPossibilities` in hopes of making more clear... not sure if I accomplished that.

Anyway, to answer your question: Yes, a card that is revealed is still considered an 'unknown card' / 'deckPossibility'.  When I attempted to add code that would subtract it from the `cardCountMap` beforehand it didn't seem to make a noticeable impact, although it's possible I had a bad implementation of that.  Changes to the `cardCountMap` need to be undone by the time we exit the reducer, since it's mutating the cached singleton map.

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: I added this as a comment to the code

Smandon1:
Alright, sorry for the delayed / two-part response to the comments.  I believe I've seen / responded to them all now at least and am now ready for the response to the responses :P.

Smandon1:
In regards to my comment "There is some reuse from action-to-action", I meant to mention what the reuse is...

When a possibility is removed from `possibleCardsForEmpathy` then that possibility will not be recalculated on future actions (since we only consider the `possibleCardsForEmpathy` from the previous action and reduce it from there).

We also won't recalculate a card on an action if the deck possibilities are the same for a card as they were the previous action (the `deckPossibilitiesDifferent` method, used under the `shouldCalculateCard` method).

timotree3:
> When a possibility is removed from possibleCardsForEmpathy then that possibility will not be recalculated on future actions (since we only consider the possibleCardsForEmpathy from the previous action and reduce it from there).

I was imagining storing a global array of deduction results where the index is the turn segment, that way when you're about to calculate turn segment #10, you check if there are any items at index < 10 and if so base the deduction off the most recent for re-use.

In any case. Let's leave that as a future possibility, especially given that you think it would worsen the testability of the code.

client/src/game/reducers/cardDeductionReducer.ts
```diff
@@ -0,0 +1,410 @@
...
+    }
+  }
+  cardCountMap[suit][rank] += 1;
+  return false;
+}
+
+// We've found a working combination of cards. If the cardCountMap is not 0 for a particular
+// possibility then that possibility is now validated and we no longer need to do a separate search
+// for it. Conversely, if the cardCountMap is 0 for a possibility, then we need to leave it in the
+// list of possibilities to validate, so we can do a separate search for it.
```
timotree3, line 367: I think this wording is clearer:
```suggestion
// We've validated a specific possibility for a specific card from a specific perspective by finding a working combination of card identities. Now we check If it's possible to swap the identity of that specific card and still have a working combination, if so, then the new identity for that card is also valid and doesn't need to be validated again.
```

I also think this comment would be more helpful to the reader if it came earlier. Would it make sense to move this up to where the `validPossibilities` variable is first defined and explain that the variable tracks this phenomenon?

client/src/game/reducers/cardsReducer.test.ts
```diff
@@ -294,63 +455,461 @@ describe("cardsReducer", () => {
...
+      let deck: readonly CardState[] = [defaultCard, secondCard];
+      // P1 draws a yellow 1
+      const gameStateDrawP2 = { ...gameState, hands: [[], [0], []] };
+      deck = cardsReducer(deck, draw(1, 0, 1, 1), gameStateDrawP2, metaData);
+
+      // P2 draws a red 5
+      const gameStateDrawP1 = { ...gameState, hands: [[], [0], [1]] };
+      deck = cardsReducer(deck, draw(2, 1, 0, 5), gameStateDrawP1, metaData);
+
+      // Expect the remaining card to not a possibility for a red 5
```
timotree3, line 619:
```suggestion
      // Expect the remaining card to not have a possibility for a red 5
```

timotree3 (approved):

timotree3:
Thanks again! @Zamiell, feel free to merge after two remaining comments are addressed

client/src/game/reducers/cardDeductionReducer.ts
Smandon1: I definitely like your wording better, especially using the word 'identity'.  I also agree it probably makes sense to move it closer to where `possibilitiesToValidate` is initially defined.  Moving it up there changed the perspective of the comment, since we're now referring to code before a (possible) working combination is found.  I tried to repurpose your wording, given the new comment location.

timotree3 (approved):

Zamiell:
thanks smandon

Smandon1:
You bet.  I'm excited :)

PR #1616, opened 2020-08-13
Cut loading times by lazy initializing

robertos:
This reduces loading times (drawing cards etc.) by more than 50%, by only constructing elements when absolutely necessary


robertos:
@Zamiell don't merge this yet, there's a small regression where cards don't animate back to the deck when going back. I'll fix it tomorrow, but in the meantime feel free to try this out :)

Zamiell:
nice thankss

PR #1624, opened 2020-08-15
Ensure the action log is shown on first open

robertos:
Repro (before this PR): open a replay, go to last turn, click action log, it's blank.


Zamiell:
thanks

PR #1629, opened 2020-08-16
Ensure the text is hidden when clue is quack

robertos:
Fix quack arrow graphical glitch (reported on discord...)


Zamiell:
ah nice, i knew it was something simple like that, thanks

PR #1634, opened 2020-08-16
Resize canvas when the window changes size

### LINKED ISSUE #212: automatic window resize functionality

robertos:
Fix #212


Zamiell:
does this actually work? the width and height is baked into everything in the 1000 line long drawUI.ts file

padiwik:
ya it works, very nice

PR #1683, opened 2020-08-25
server: shared missing scores: Fix operand order for IsBetterThan invocation

timotree3:
loosely tested locally


Zamiell:
thanks

Issue #1696, opened 2020-08-27
watching game replay from JSON borked


timotree3:
I opened the replay from JSON data of game `#279627` and all the cards were either red or yellow when in the real game there were tons of other suits. Due to the lack of green cards in the opening hand of the replay, the rest of the replay is borked because the first clue of the game (greens) is "empty".

Steps to reproduce:
1. go to https://hanab.live/export/279627
2. watch a JSON replay with that data

Zamiell:
fixed

PR #1699, opened 2020-08-27
client: card status: Update visisble card status whenever a clue is given

### LINKED ISSUE #1681: Cluing trash doesn't make it opaque
https://hanab.live/replay/279627/39 the fill should reload to opacity 1 as soon as the clue happens

timotree3:
Should fix #1681. I haven't tested this because playing a game locally isn't working well and due to #1696 I can't open the test case replay

### DIFF
--- a/client/src/game/ui/reactive/view/cardsView.ts
+++ b/client/src/game/ui/reactive/view/cardsView.ts
@@ -122,10 +122,6 @@ const subscribeToCardChanges = (order: number) => {
       location: card.location,
     };
   }, () => updateCardStatus(order));
-  sub((c) => ({
-    numPositiveClues: c.numPositiveClues,
-    location: c.location,
-  }), () => updateBorder(order));
 
   // Pips
   sub((c) => ({
--- a/client/src/game/ui/reactive/view/cardsView.ts
+++ b/client/src/game/ui/reactive/view/cardsView.ts
@@ -113,7 +113,7 @@ const subscribeToCardChanges = (order: number) => {
   // Card fade and critical indicator
   subFullState((s) => {
     const card = s.visibleState!.deck[order];
-    const status = card.suitIndex !== null && card.rank !== null
+    const status = (card.suitIndex !== null && card.rank !== null)
       ? s.visibleState!.cardStatus[card.suitIndex!][card.rank!]
       : null;
     return {
--- a/client/src/game/ui/reactive/view/cardsView.ts
+++ b/client/src/game/ui/reactive/view/cardsView.ts
@@ -114,7 +114,7 @@ const subscribeToCardChanges = (order: number) => {
   subFullState((s) => {
     const card = s.visibleState!.deck[order];
     const status = (card.suitIndex !== null && card.rank !== null)
-      ? s.visibleState!.cardStatus[card.suitIndex!][card.rank!]
+      ? s.visibleState!.cardStatus[card.suitIndex][card.rank]
       : null;
     return {
       status,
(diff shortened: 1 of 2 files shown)


timotree3:
An alternative solution/follow-up is being discussed in #1681 that I would happily implement.

timotree3:
Okay switched to that other solution. I did locally test this and the client seems to work fine although I can't reproduce the bug on the old revision either.

client/src/game/ui/reactive/view/cardsView.ts
```diff
@@ -110,6 +110,23 @@ const subscribeToCardChanges = (order: number) => {
     location: c.location,
   }), () => updateBorder(order));
 
+  // Card fade and critical indicator
+  subFullState((s) => {
+    const card = s.visibleState!.deck[order];
+    const status = card.suitIndex !== null && card.rank !== null
```
robertos, line 116: parens around the condition because the operator precedence of ? is not clear

timotree3: if it's the other grouping than we're &&-ing with a non boolean. I don't think that's possible

robertos: You can definitely && a non-boolean - welcome to javascript :/ Everything casts to everything silently

client/src/game/ui/reactive/view/cardsView.ts
```diff
@@ -110,6 +110,23 @@ const subscribeToCardChanges = (order: number) => {
     location: c.location,
   }), () => updateBorder(order));
 
+  // Card fade and critical indicator
+  subFullState((s) => {
+    const card = s.visibleState!.deck[order];
+    const status = (card.suitIndex !== null && card.rank !== null)
+      ? s.visibleState!.cardStatus[card.suitIndex!][card.rank!]
```
Zamiell, line 117: why put the bangs on suitindex and rank?

timotree3: I didn't know how to address the nullability besides this approach. I'm new to TS so any suggestion welcome

client/src/game/ui/reactive/view/cardsView.ts
timotree3: fair. fixed

client/src/game/ui/reactive/view/cardsView.ts
timotree3: Wow I guess I can just remove them. What kind of sorcery is typescript doing??

client/src/game/ui/reactive/view/cardsView.ts
```diff
@@ -110,6 +110,19 @@ const subscribeToCardChanges = (order: number) => {
     location: c.location,
   }), () => updateBorder(order));
 
+  // Card fade and critical indicator
+  subFullState((s) => {
```
Zamiell, line 114: if it only uses visible state, then shouldnt the sub be to the visible state?

robertos: It's literally the same

robertos: the "sub visible" is a shortcut for what he did

Zamiell: if its the same, then use the provided helper and put the code in the correct section

robertos: The helper is not in this file.

Zamiell: ok never mind then

Zamiell:
thanks tim for looking into this

PR #1715, opened 2020-08-30
style: 'one less' should be 'one fewer'

aliblong:


padiwik:
Doesn't "one less card" sound better in this context? It is also used more in google results vs "one fewer card"

aliblong:
> Doesn't "one less card" sound better in this context? It is also used more in google results vs "one fewer card"

Sound better to whom? "Fewer" sounds correct to me, a native English speaker, and matches the traditional prescription ("cards" is a countable noun).

Zamiell:
one less card sounds better to me

PR #1728, opened 2020-09-02
Saner modal backgrounds

### LINKED ISSUE #1026: Escape key does not close warnings/errors
When a warning/error pops up on the screen, it should be possible to close it with the keyboard like other elements of the UI. However, it is not focused which means the Enter key will act on the chat rather than the warning/error. The Escape key reverts the dimming of the background but does not close the warning/error.

padiwik:
* Fix #1026
* Keep chat modal open (but I can't figure out a nice way to keep the password box or create game dialog open without looking like there is lag in the no-warning case. perhaps a variable that tracks lastOpenTooltip??)
* Use a solid color instead of letting the wood background bleed through (to undo this, make page-cover have the same properties as body)
* Don't allow clicking on background elements (clicking the background still closes tooltips, not sure why, but we want that)
* Yay for functionizing the code

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -64,4 +64,7 @@
   "python.formatting.provider": "black",
   "python.linting.pylintEnabled": true,
   "python.linting.enabled": true,
+  "files.exclude": {
+    "**/misc/keldon/**": true
+  },
 }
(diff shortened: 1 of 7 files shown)


Zamiell:
lgtm

Issue #1739, opened 2020-09-12
feature: Identity list notes affect pips


timotree3:
Parse notes that are a comma separated series of identities and show them in the UI.

For example, if a card had the note `y4,m4` it should morph into a 4 and it's color pips should be yellow and multi.

I believe that just like identity notes, these should be ignored in empathy.

timotree3:
Bonus would be to parse the common shorthand for identity lists that would make `r234` expand to `r2,r3,r4` and `ry23` expand to `r2,r3,m2,m3`

timotree3:
I want to work on this tomorrow

padiwik:
personally i like the shorthand `rm23` because it's quicker to parse in my head. I also think it's a good compromise to keep it for the people who don't like note morphing cards.

timotree3:
> personally i like the shorthand `rm23` because it's quicker to parse in my head.

Me too. To be clear, I mean that we would parse that shorthand as though it expanded to the comma-separated list, I don't actually want it to expand that way in the note. 

> I also think it's a good compromise to keep it for the people who don't like note morphing cards.

Are you saying that you know people that would prefer not to have this feature?

Zamiell:
> Are you saying that you know people that would prefer not to have this feature?

i think that is the case

you are trying to combine two disparate features, which actually removes information from the player.
meaning - a player can hover over a note to see the possibilities from the note, and they can look at the pips to see the explicit information / clues.
in your implementation, if they want to see the latter, they would have to delete the note and then readd the note, which probably isn't desired behavior.

timotree3:
> in your implementation, if they want to see the latter, they would have to delete the note and then readd the note, which probably isn't desired behavior.
No they wouldn't. I think it would work like identity notes do. As soon as you hold left-click on the card or drag a card it would show the card with notes ignored

Zamiell:
oh, then that is probably okay

Zamiell:
status on this?

timotree3:
will resume work once I have a working dev environment.

timotree3:
I'll work on this today

PR #1741, opened 2020-09-12
client: Improve note parsing logic

timotree3:
This makes the note parsing logic do far less work. This changes behavior in two ways:

- Partial identity notes can now combine, for example `[r] [1]` now produces a morphed red 1
- Abbreviation identity notes can now have a space in the middle, for example `r 1` now produces a morphed red 1

I have tested this locally extensively

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -8,6 +8,7 @@ module.exports = {
   env: {
     browser: true,
     jquery: true,
+    es2020: true,
   },
 
   // We need to specify some additional settings in order to make the linter work with TypeScript:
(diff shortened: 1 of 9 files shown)


timotree3:
I plan to build on this in my implementation of #1739

client/.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
   env: {
     browser: true,
     jquery: true,
+    es2020: true,
```
Zamiell, line 11: why

timotree3: regex matchAll

timotree3: Note that I don't think this changes our browser compatibility because the target is still es2018. I imagine matchAll is getting expanded to es2018 compatible code

Zamiell:
thanks

client/.eslintrc.js
timotree3: Actually I don't think this is true. I'll make a PR that switched away from using matchAll

PR #1744, opened 2020-09-12
client: Have finesse border take precedence over chop move border

### LINKED ISSUE #1742: Finesse note should take precedence over chop move note

timotree3:
Fixes #1742


Zamiell:
thanks

Issue #1710, opened 2020-08-29
Button to change variant in pregame


padiwik:
so that people can choose from a dropdown, since /setvariant is very annoying to deal with (you need to know the exact variant name and type it in exactly)

PR #1758, opened 2020-09-19
Change variant tooltip in pre-game

vEnhance:
Implement #1710 because I want this feature too :stuck_out_tongue:


client/src/lobby/nav.ts
```diff
@@ -94,10 +146,11 @@ export const init = () => {
 
 const initTooltips = () => {
   const tooltips = [
-    'create-game',
-    'replay',
-    'resources',
-    'settings',
```
Zamiell, line 100: why are you renaming these

client/src/lobby/nav.ts
```diff
@@ -57,6 +58,57 @@ export const init = () => {
...
+      if (event.key === 'Enter') {
+        event.preventDefault();
+        $('#change-variant-update').click();
+      }
+    });
+
+    // Update button trigger
+    $('#change-variant-update').on('click', () => {
+      const variantName = ($('#change-variant-dropdown').val() as string).trim();
+      console.log(variantName);
```
Zamiell, line 97: why is this here

vEnhance: mb

client/src/lobby/nav.ts
vEnhance: I wanted to `initToolTips` also hit `nav-buttons-pregame-change-variant`

client/src/lobby/pregame.ts
```diff
@@ -423,7 +423,7 @@ const drawPlayerBox = (i: number) => {
 };
 
 export const enableStartGameButton = () => {
-  // Enable or disable the "Start Game" button,
+  // Enable or disable the "Start Game" and "Change Variant" button,
   // depending on if we are the game owner and enough players have joined
```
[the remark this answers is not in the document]
vEnhance: The dropdown is okay with just one player (leader) in pregame, and the code is written that way.

I can see the comment might be a little misleading as written --- feel free to reword as you see fit.

vEnhance:
Anything else I should do for this PR?

client/src/lobby/nav.ts
Zamiell: using a `games-` prefix implies that they are part of the game screen, when they are not
why not use a `lobby-` prefix

vEnhance: The `games-` prefix isn't by me, was there before, but I can rename them all to `lobby-`. Agree it makes more sense.

client/src/lobby/pregame.ts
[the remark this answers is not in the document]
vEnhance: 41ded0a clarifies the comment

client/src/lobby/nav.ts
vEnhance: 18ce2e41dcfebb2467ee20a4cf317c861452502a does this renaming. this is a fairly deep change... hope it doesn't break anything (though it seems to work as far as my testing can see)

Zamiell:
> Anything else I should do for this PR?

1) make the button the default css, e.g. black
2) center the button in the tooltip

PR #1764, opened 2020-09-23
Vertically flip prism pip

timotree3:
In order to make it look less like the yellow pip.


timotree3:
Before:
Prism Before
After:
Prism After

Zamiell:
cool, thanks

PR #1766, opened 2020-09-24
Fix DDA notification

### LINKED ISSUE #1765: #305547 DD yellow border does not appear

timotree3:
Fixes #1765


timotree3:
Diagnosis: DDA notification isn't working whenever the player has no clued cards in hand. Test case: https://hanab.live/replay/305547/12

It doesn't show yellow border currently, and on my branch it shows it

Zamiell:
nice thank you

PR #1778, opened 2020-10-01
client: Fix note keyword regex

timotree3:


client/src/game/ui/notes.ts
```diff
@@ -75,7 +75,7 @@ export const set = (order: number, note: string) => {
 const getNoteKeywords = (
   note: string,
 ) => {
-  const regexp = /\[([^\]]*)\]|\|([^|[]*)$|^([^|]+)$/g;
+  const regexp = /\[(.*?)\]|\|([^[|]*$)|(^[^[|]+$)/g;
```
Zamiell, line 78: can you add a comment explaining what the regex is doing piece by piece

timotree3: does what I added help?

Zamiell: perfect, thank you very much

i always like to write healthy comments next to any non-trivial code like this

PR #1779, opened 2020-10-01
Bugfix for pre-discard

padiwik:
i don't think anything else got borked in the refactor 4eaea506ba159d27e6a024dcaedc74ab45d1acaf


Zamiell:
nice thanks

PR #1780, opened 2020-10-01
Fix for buggy clue arrows

### LINKED ISSUE #1774: Clue arrows can glitch onto the wrong cards
Seems to happen when moving quickly in a replay, as if the animation of the arrow becomes refocused on the wrong turn's cards.

padiwik:
Fix #1774 

I believe this has always been a bug, but it's more noticeable now because of the reverse play animations


Zamiell:
ok, we can try this out

padiwik:
another idea that might work:

`if arrow already animating: return`

but then it might have weird behavior like showing only 2/3 of the previous turn's arrows

PR #1782, opened 2020-10-01
Ignore chop moved notes w.r.t card fade during postgame

### LINKED ISSUE #1781: Chop moved trash cards should still fade during postgame
They currently remain opaque

timotree3:
Fixes #1781


timotree3:
does the change to the blank part actually make sense? I don't understand why we want a blank note to affect fading to begin with

Zamiell:
> does the change to the blank part actually make sense?

yes

> I don't understand why we want a blank note to affect fading to begin with

it affects fading because you dont want a blank card to be faded

Zamiell:
thanks

PR #1784, opened 2020-10-05
Update example_game_with_comments.jsonc

akhti:
Make the file loadable (after removing comments)
Fix bugs in the comments
Make deck have 50 cards

I still can only replay the first move, but at least it shows the table


Zamiell:
thx

Zamiell:
you're from FB?

akhti:
> you're from FB?

yup

Zamiell:
cool thanks

PR #1808, opened 2020-10-31
Fix #1807

### LINKED ISSUE #1807: no yellow border on #331538 t20

timotree3:


Zamiell:
what was the bug?

timotree3:
It was checking whether the hand of the player who discarded the card was fully clued, not the player who is taking a turn afterwards. e.g. in the linked replay kimbi's hand was all clued minus the card she just discarded, but what mattered was whether Zam's hand was all clued (which it wasn't)

Zamiell:
ah, very nice, thanks for fixing this!

PR #1813, opened 2020-11-01
Future eff tests

timotree3:


timotree3:
hmm it looks glitchy as you use it. as you discard, the future efficiency flickers down a bit and then up, maybe it is lowering the pace by one, updating the efficiency, and then gaining a clue and updating it again?

Zamiell:
no idea, start putting in breakpoints or something

Zamiell:
or put in a shitload of console.log everywhere

Zamiell:
[ ] add tests for all of the interesting stats functions besides doubleDiscard
[ ] fix pace with contrarian
[ ] fix pace + efficiency with all or nothing

timotree3:
> add tests for all of the interesting stats functions besides doubleDiscard
this is harder than I thought because everything is intermingled with statsReducer. Maybe I should write tests for statsreducer?

Zamiell:
not sure what you mean - there are already integration tests, add more of those if you want?

timotree3:
well to do those in the same style I would have to generate a game JSON that would trigger the past bug

Zamiell:
add the file statsReducer.test.ts and add tests to it i guess

Zamiell:
actually no because it needs an entire game state, so you cant test the individual functions

Zamiell:
so you need to add something to integration.test.ts

and no you dont need to use a JSON, you can just code yourself a simple game state from scratch

Zamiell:
actually

Zamiell:
i dont actually know

Zamiell:
look at how turnReducer.test.ts works, it is a similar thing as statsReducer - both .turn and .stats are sub chunks of the bigger state

Zamiell:
i think you can just copy that to do what you want

Zamiell:
e.g.

```
      const drawAction = draw(0, 0, 0, 1); // Draw a red 1
      state = turnReducer(state, drawAction, defaultGameState, defaultMetadata);

      for (let i = 0; i < 3; i++) {
        const playAction = play(0, i, 0, 1); // Play the last red 1 that was drawn
        state = turnReducer(
          state,
          playAction,
          defaultGameState,
          defaultMetadata,
        );
        const drawAction2 = draw(0, i + 1, 0, 1); // Draw another red 1
        state = turnReducer(
          state,
          drawAction2,
          defaultGameState,
          defaultMetadata,
        );
      }
```

timotree3:
Just debugged. future efficiency number dips easier for a moment until the draw action reduces and it goes back to normal every time that a play or discard occurs

timotree3:
Why doesn't that happen the pace?

timotree3:
Debugged the pace and the pace also increase for a moment, it even calls the function that changes the text on the label, and then calls `batchDraw`, but nothing on screen changes and it has time to set it back to normal and call `batchDraw` again. Maybe the problem is the way that `batchDraw` was working? It was drawing part-way through the process and therefore showed a wrong efficiency?

timotree3:
That seems like an unrelated bug so we can probably merge this. The solution to that bug would probably be to change the time at which `cardsRemainingInTheDeck` is changed such that it changes during a `play` or `discard` action instead of during a `draw` action.

timotree3:
another bug found

timotree3:
That one only occurs once a 5 has been played so didn't show up in any tests.

Zamiell:
[ ] add tests for all of the interesting stats functions besides doubleDiscard
[ ] fix pace with contrarian
[ ] fix pace + efficiency with all or nothing

Zamiell:
is #1 done?

timotree3:
No I haven't done those tests and don't want to anymore. I also don't care about all or nothing because that variant is way too easy. I would fix pace with contrarian but I don't see why that should hold up this PR. I'll do that later today as a new PR unless this one hasn't been merged yet.

timotree3:
Okay. added contrarian support.

Zamiell:
looks good thanks!

PR #1823, opened 2020-11-04
fix upper arms in the shruggy

adherr:


Zamiell:
who are u?

adherr:
dorkman

adherr:
I don't golang. But I markdown!

Zamiell:
not familiar with that name, but thanks!

PR #1827, opened 2020-11-09
Identity list note

### LINKED ISSUE #1739: feature: Identity list notes affect pips

timotree3:
This sets up parsing for identity list notes and causes a note like `g2, r2` to morph the card into a 2.

Resolves #1739 

TODO:
- [x] Parse shorthand. (e.g `gr2`) EDIT: let's leave this for another PR
- [x] Change pips corresponding to such a note in addition to the aforementioned morphing. (e.g. `g2,r2` produces a morphed 2 with a green pip and a red pip.)

### DIFF
--- a/client/src/game/types/CardNote.ts
+++ b/client/src/game/types/CardNote.ts
@@ -1,5 +1,5 @@
 export default interface CardNote {
-  // The possible card identities included in the note. Empty if none.
+  // The possible card identities included in the note (empty if none)
   possibilities: Array<[number, number]>;
   readonly knownTrash: boolean;
   readonly needsFix: boolean;
--- a/client/src/game/types/CardNote.ts
+++ b/client/src/game/types/CardNote.ts
@@ -1,8 +1,6 @@
 export default interface CardNote {
-  // The suit corresponding to the note written on the card, if any
-  suitIndex: number | null;
-  // The rank corresponding to the note written on the card, if any
-  rank: number | null;
+  // The possible card identities included in the note. Empty if none.
+  possibilities: Array<[number, number]>;
   readonly knownTrash: boolean;
   readonly needsFix: boolean;
   readonly chopMoved: boolean;
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -66,8 +66,7 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   private wrench: Konva.Image;
 
   note: CardNote = {
-    suitIndex: null,
-    rank: null,
+    possibilities: [],
     chopMoved: false,
     needsFix: false,
     knownTrash: false,
@@ -357,11 +356,18 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
       );
     }
 
-    // If we have a note identity on the card, show the suit corresponding to the note
-    if (this.note.suitIndex !== null) {
-      return this.variant.suits[this.note.suitIndex];
+    // If we have a note on the card and it only provides possibilities of the same suit,
+    // show that suit
+    if (this.note.possibilities.length !== 0) {
+      const [candidateSuitIndex] = this.note.possibilities[0];
+      if (
+        this.note.possibilities.every(
+          ([suitIndex]) => suitIndex === candidateSuitIndex,
+        )
+      ) {
+        return this.variant.suits[candidateSuitIndex];
+      }
     }
-
     return unknownSuit;
   }
 
@@ -375,15 +381,24 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
       return UNKNOWN_CARD_RANK;
     }
 
-    // If we have a note identity on the card, show the rank corresponding to the note
-    // (specifically for stack bases in ongoing games; we want notes to have precedence in this case
-    // so that players can make notes in "Throw It in a Hole" variants)
+    // If we have a note on the card, and it only provides possibilities of the same rank,
+    // show that rank. (specifically for stack bases in ongoing games; we want notes to have
+    // precedence in this case so that players can make notes in "Throw It in a Hole" variants)
+
+    let noteRank = null;
+    if (this.note.possibilities.length !== 0) {
+      const [, candidateRank] = this.note.possibilities[0];
+      if (this.note.possibilities.every(([, rank]) => rank === candidateRank)) {
+        noteRank = candidateRank;
+      }
+    }
+
     if (
-      this.note.rank !== null &&
+      noteRank !== null &&
       this.state.rank === STACK_BASE_RANK &&
       !globals.state.finished
     ) {
-      return this.note.rank;
+      return noteRank;
     }
 
     // Show the rank if it is known
@@ -392,8 +407,8 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
     }
 
     // If we have a note identity on the card, show the rank corresponding to the note
-    if (this.note.rank !== null) {
-      return this.note.rank;
+    if (noteRank !== null) {
+      return noteRank;
     }
 
     return UNKNOWN_CARD_RANK;
@@ -1093,11 +1108,13 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
     // eliminated, unmorph the card
     // (e.g. a note of "r1" is now impossible because red 1 has 0 cards left)
     if (
-      !cardRules.canPossiblyBe(this.state, this.note.suitIndex, this.note.rank)
+      this.note.possibilities.every(
+        ([suitIndex, rank]) =>
+          !cardRules.canPossiblyBe(this.state, suitIndex, rank),
+      )
     ) {
       // Unmorph
-      this.note.suitIndex = null;
-      this.note.rank = null;
+      this.note.possibilities = [];
       this.setBareImage();
     }
   }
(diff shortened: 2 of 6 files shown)


client/src/game/ui/HanabiCard.ts
```diff
@@ -375,15 +381,24 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
       return UNKNOWN_CARD_RANK;
     }
 
-    // If we have a note identity on the card, show the rank corresponding to the note
-    // (specifically for stack bases in ongoing games; we want notes to have precedence in this case
-    // so that players can make notes in "Throw It in a Hole" variants)
+    // If we have a note on the card, and it only provides possibilities of the same rank,
+    // show that rank. (specifically for stack bases in ongoing games; we want notes to have
```
Zamiell, line 385: punctuation

client/src/game/ui/notes.ts
```diff
@@ -175,10 +174,62 @@ function parseRank(rankText: string): number {
...
+function getPossibilitiesFromKeyword(
   variant: Variant,
   keyword: string,
-): CardIdentity {
+): Array<[number, number]> | null {
+  const possibilities: Array<[number, number]> = [];
+  for (const substring of keyword.split(",")) {
+    const identity = parseIdentity(variant, substring);
+    if (identity.suitIndex !== null && identity.rank !== null) {
+      // encountered an identity item, add it
```
Zamiell, line 185: capitalize

client/src/game/ui/notes.ts
```diff
@@ -175,10 +174,62 @@ function parseRank(rankText: string): number {
...
+          possibilities.findIndex(
+            (possibility) =>
+              possibility[0] === identity.suitIndex && possibility[1] === rank,
+          ) === -1
+        ) {
+          possibilities.push([identity.suitIndex, rank]);
+        }
+      }
+    } else if (identity.suitIndex === null && identity.rank !== null) {
+      // encountered a rank item, expand to all cards of that rank.
```
Zamiell, line 211: both problems

timotree3:
image
:tada:

timotree3:
Feel free to try this out locally and give any feedback. It seems to work perfectly now.

client/src/game/types/CardNote.ts
```diff
@@ -1,8 +1,6 @@
 export default interface CardNote {
-  // The suit corresponding to the note written on the card, if any
-  suitIndex: number | null;
-  // The rank corresponding to the note written on the card, if any
-  rank: number | null;
+  // The possible card identities included in the note. Empty if none.
```
Zamiell, line 2: punctatuion

timotree3: :angry:

client/src/game/ui/HanabiCardMouse.ts
```diff
@@ -195,13 +195,17 @@ function shouldShowLookCursor(card: HanabiCard) {
...
   // For ongoing games, only show the cursor for our hand if it has a custom card identity
-  if (
-    (card.note.suitIndex !== null &&
-      card.note.suitIndex !== card.state.suitIndex) ||
-    (card.note.rank !== null && card.note.rank !== card.state.rank) ||
-    card.note.blank ||
-    card.note.unclued
-  ) {
+
+  // Check if there exists a possibility from clues that the note declares impossible.
```
Zamiell, line 199: punctuation

client/src/game/ui/HanabiCard.ts
```diff
@@ -375,15 +388,35 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
...
+    let noteRank = null;
+    if (this.note.possibilities.length !== 0) {
+      const possibleCardsFromNoteAndClues = this.note.possibilities.filter(
+        ([suitIndexA, rankA]) =>
+          this.state.possibleCardsFromClues.findIndex(
+            ([suitIndexB, rankB]) =>
+              suitIndexA === suitIndexB && rankA === rankB,
+          ) !== -1,
+      );
+      const [, candidateRank] = possibleCardsFromNoteAndClues[0];
```
Zamiell, line 404: i'm not a fan of this notation, it seems confusing and hard to read
how about [0][1]

client/src/game/ui/HanabiCardMouse.ts
timotree3: you're killing me <3

client/src/game/ui/HanabiCard.ts
```diff
@@ -375,15 +388,35 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
...
+      if (
+        possibleCardsFromNoteAndClues.every(
+          ([, rank]) => rank === candidateRank,
```
Zamiell, line 407: same here

client/src/game/ui/notes.ts
```diff
@@ -175,10 +174,62 @@ function parseRank(rankText: string): number {
...
+          possibilities.findIndex(
+            (possibility) =>
+              possibility[0] === suitIndex && possibility[1] === identity.rank,
+          ) === -1
+        ) {
+          possibilities.push([suitIndex, identity.rank]);
+        }
+      }
+    } else {
+      // Encountered invalid identity. do not parse keyword as an identity list
```
Zamiell, line 224: .

timotree3: What change do you want here?

Zamiell: semi colon

client/src/game/ui/notes.ts
```diff
@@ -196,110 +247,79 @@ export function getIdentityFromKeyword(
...
-  if (note.suitIndex === null && note.rank !== null) {
-    // Only the rank was specified
-    modals.warningShow(`That card cannot possibly be a ${impossibleRank}.`);
-    note.rank = null;
-    return;
-  }
-
-  if (note.suitIndex !== null && note.rank !== null) {
-    // Both the suit and the rank were specified
+  // We have specified a list of identities where none are possible.
```
Zamiell, line 308: .

client/src/game/ui/HanabiCard.ts
timotree3: I like it better. It's put a name to the variable rather than `[0]` which you'd have to find where the value is produced to understand.

client/src/game/ui/HanabiCardMouse.ts
Zamiell: if you would do it right from the beginning then it wouldn't be my fault =p

client/src/game/ui/HanabiCard.ts
Zamiell: `const candidateRank = possibleCardsFromNoteAndClues[0][1];`

this is what i meant

client/src/game/ui/HanabiCard.ts
timotree3: unlike the previous case this cannot be replacing with a variable, thus changing this to indexing would remove information from the reader

timotree3: do you insist?

Zamiell: > unlike the previous case this cannot be replaced with a variable, thus changing this to indexing would remove information from the reader
then by all means keep it as written = )

client/src/game/ui/HanabiCard.ts
timotree3: should I change this still or keep both the same?

Zamiell: change please.
(don't you agree that having one index be on the left side and one index be on the right side is just bad code?)

timotree3: > change please.
K

> (don't you agree that having one index be on the left side and one index be on the right side is just bad code?)
No. The two indexes mean different things. One of them is an index into an array. The other one is essentially a field access on a tuple of (suitIndex, rank). I would rather access a field by name than by number.

Zamiell: you can use array destructuring for all kinds of arrays, not just tuples.
so the line doesn't necessarily betray that the data structure is a tuple.
i guess your point is that by using array destructuring, you inject-into-the-code-style an implication that a tuple exists?
it just comes off to me like using arcane syntax for not a huge benefit

timotree3: In the language I'm used to, doing the destructuring would only be valid if the array was known to have exactly two elements and if that were the case in TS, then doing the destructuring would make it very clear to the reader that it is not in fact an array. Knowing that in TS you can destructure even if it has more elements, I don't really care. I'll just change both to indexing I suppose

Zamiell: > In the language I'm used to, doing the destructuring would only be valid if the array was known to have exactly two elements
indeed, that sounds pretty useful. what language is that

timotree3: Rust

Zamiell: and by the way, array destructuring is a JS feature not a TS feature

Zamiell:
thanks tim

PR #1851, opened 2020-11-19
Fix statsReducer to only be concerned with relevant action types

timotree3:
and consequently fix DDA sounds.

Still testing locally.

### DIFF
--- a/client/src/game/reducers/statsReducer.ts
+++ b/client/src/game/reducers/statsReducer.ts
@@ -2,6 +2,7 @@
 // as a result of each action
 
 import produce, { Draft } from "immer";
+import { ensureAllCases } from "../../misc";
 import { getVariant } from "../data/gameData";
 import { clueTokensRules, turnRules, variantRules } from "../rules";
 import * as statsRules from "../rules/stats";
@@ -28,6 +29,7 @@ function statsReducerFunction(
     case "clue": {
       // A clue was spent
       stats.potentialCluesLost += 1;
+      stats.doubleDiscard = false;
 
       break;
     }
@@ -54,14 +56,37 @@ function statsReducerFunction(
         // similar to what happens when the team gets a strike
         stats.potentialCluesLost += clueTokensRules.discardValue(variant);
       }
+      stats.doubleDiscard = false;
 
       break;
     }
+
+    case "discard": {
+      // Handle double discard calculation
+      stats.doubleDiscard = statsRules.doubleDiscard(
+        action.order,
+        currentState,
+        variant,
+      );
+      break;
+    }
+
+    case "draw":
+    case "cardIdentity":
+    case "gameOver":
+    case "turn":
+    case "playerTimes":
+      return;
 
     default: {
-      break;
+      ensureAllCases(action);
+      return;
     }
   }
+
+  const actionDrewCard = originalState.cardsRemainingInTheDeck >= 1;
+  const deckSize =
+    originalState.cardsRemainingInTheDeck - (actionDrewCard ? 1 : 0);
 
   // Handle max score calculation
   if (action.type === "play" || action.type === "discard") {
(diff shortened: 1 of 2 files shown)


timotree3:
Ready for review

timotree3:
Bug: Game over sounds no longer play

timotree3:
Actually the bug is that game over doesn't register at all. It shows player times but the sound doesn't play and the client doesn't automatically switch to shared replay

Zamiell:
add a comment of
```
// Various stat calculation is below; this code only runs on play, discard, clue, or strike actions
```

timotree3:
It's there

Zamiell:
how come it doesn't show in https://github.com/Zamiell/hanabi-live/pull/1851/files

Zamiell:
oh it does now

Zamiell:
so why isn't game over working

timotree3:
no idea

Zamiell:
is it broken on master or just in the pr?

timotree3:
just in PR. Also, terminate button does go to shared replay but plays the wrong sound

Zamiell:
so statsreducer probably needs to know about gameover action?

timotree3:
hahahahaha. `getSoundType` is only ever called in `statsReducer`. that explains it. I assumed it would be called from somewhere else

timotree3:
Okay now the right sound plays but it still doesn't enter the shared replay

timotree3:
Where is the code that makes the client enter shared replay after a game?

Zamiell:
gameCommands.ts line 81
which in turn calls the code in stateReducer.ts line 61

timotree3:
This is the error when terminating:

```
Uncaught TypeError: Cannot perform 'get' on a proxy that has been revoked
    at equal (index.js:11)
    at equal (index.js:38)
    at observeStore.ts:41
    at Array.filter (<anonymous>)
    at handleChange (observeStore.ts:30)
    at Object.dispatch (redux.js:222)
    at gameCommands.ts:413
    at Object.spectators (websocketInit.ts:100)
    at Connection.onMessage (Connection.ts:52)
```
Any idea what would cause that?

Zamiell:
its because you can't modify a thing in redux

timotree3:
so somewhere in `statsReducer` I added code which modified something illegally?

Zamiell:
i forgot the specifics tbh, i dont understand redux at all

timotree3:
Should be fixed now. That was so annoying

Zamiell:
what was the fix

timotree3:
See latest commit

Zamiell:
why was it being recorded there and will removing it cause any bugs?

client/src/game/reducers/statsReducer.ts
```diff
@@ -54,15 +56,52 @@ function statsReducerFunction(
...
         stats.potentialCluesLost += clueTokensRules.discardValue(variant);
       }
+      stats.doubleDiscard = false;
 
       break;
     }
 
-    default: {
+    case "discard": {
+      // Handle double discard calculation
```
Zamiell, line 65: isn't this a bug because double discard will never be set to false unless two people discard in a row

timotree3: notice the other places in the code where `stats.doubleDiscard = false` is written: lines 32 and 59

timotree3: although I'm not sure if this accounts for dda situations caused by a misplay. the previous code didn't seem to either and yet I've seen it work. Is that because a misplay is encoded as a "strike" action plus a "discard" action?

Zamiell: no idea, strike action will be removed at some point so if it relies on a strike action then that is a bug

timotree3: the only thing here that relies on strike action is `potentialCluesLost` :P

PR #1854, opened 2020-11-20
Tiny tweaks to clean up the ui (mostly keldon mode)

Hyphen-ated:
in keldon mode:
* move hands and labels around so things don't overlap or go offscreen (which they had been doing because of the black bar addition a while ago)
* make the black bar a roundrect, like every other rect on the website.

also, reduce the font size of the replay ID text on the deck by 5%, so that when it says "JSON replay" it no longer runs riiiiiight up to the edge, which was ugly.


Zamiell:
thanks

PR #1872, opened 2020-11-21
In game hypo

### LINKED ISSUE #1156: in-game hypothetical

Zamiell:
Resolves #1156


Zamiell:
opening this now because i want to see the diff of the whole branch

client/src/game/ui/drawReplayArea.ts
```diff
@@ -220,18 +220,22 @@ export default function drawReplayArea(winW: number, winH: number): void {
...
+  // The "Exit Replay" button
   globals.elements.replayExitButton = new Button({
-    x:
-      (replayButtonValues.x +
-        replayButtonValues.w +
-        replayButtonValues.spacing / 2) *
-      winW,
+    x: bottomLeftReplayButtonValues.x * winW,
     y: bottomButtonValues.y * winH,
-    width: (replayButtonValues.w * 2 + replayButtonValues.spacing * 2) * winW,
```
timotree3, line 234: I didn't change this on purpose

timotree3: oh I see that was you.

Zamiell: ?

timotree3:
The feature seems to be working smoothly.

TODO:
- [ ] prompt user for card identity when playing/discarding an unknown card
- [ ] allow clues to self when cards are morphed and would be touched
- [ ] check if you rewind start a hypo and draw a card that is fully identified since you see it in the present everything still works
- [ ] desirable? - add back Show Drawn Cards button or change default to show drawn cards for the above-mentioned case (removed because it conflicts with Your Turn indicator)
- [ ] automatically morph all cards in hand according to identity notes when starting hypothetical for ease of use.
- [ ] test that shared replay still works
- [ ] fix enter hypothetical button in private post-game replay

Zamiell:
> add back Show Drawn Cards button

how could that be possible because you dont know what drawn cards will be

Zamiell:
the rest sound good

timotree3:
it only shows cards that have already been drawn in the present

Zamiell:
oh

thats a bit confusing, because 99% of the time people will be initiating a hypo from the present turn
thus, i dont think we need to add additional UI elements to facilitate initiating a hypo from a previous turn

timotree3:
> 99% of the time people will be initiating a hypo from the present turn
maybe you would start from a past turn to see if anybody would've acted different had your held card been X?

Zamiell:
that's true

Zamiell:
ok, so show the button, but only if the hypo start turn is not the present turn

timotree3:
sounds good. although it's also possible for the bottom card in the deck to be fully identified if your hand is locked

Zamiell:
true but we probably don't care about that

client/src/game/reducers/initialStates/initialState.ts
```diff
@@ -25,7 +38,10 @@ export default function initialState(metadata: GameMetadata): State {
     datetimeFinished: null,
 
     metadata,
-    cardIdentities: [],
+    cardIdentities: initArray(startingDeckSize, {
+      suitIndex: null,
+      rank: null,
+    }),
```
timotree3, line 44: 1. `startingDeckSize` returns the size of the deck on turn 1. Do you mean to use the size of the deck before cards are dealt?

2. Now everywhere where we checked for undefined before we need to check for {null, null}, I can update that.

3. If this is actually a good idea then there should be similar code in `loadGameJSON.ts`

Zamiell: 1) yes
2) yes
3) yes

timotree3: fixed

timotree3:
I recently learned that you can be in a shared replay and a game at the same time. Are there gonna be any bugs where we expect `shared` to equal `null` during in-game hypo but it actually doesn't because the user is doing both at once?

timotree3:
Also this PR is ready for review and merge

Zamiell:
what do you mean "at the same time"

Zamiell:
you can be on the server, but not the client

Zamiell:
fantastic work, thank you

PR #1873, opened 2020-11-21
remove deck tooltip useless text, always show icon

Hyphen-ated:


client/src/game/ui/drawUI.ts
```diff
@@ -648,7 +648,7 @@ function drawDeck() {
       y: 0,
     },
     shadowOpacity: 0.9,
-    visible: globals.state.finished && globals.state.replay.databaseID !== null,
+    visible: true,
```
Zamiell, line 651: this line isn't needed, things are visible by default

Zamiell:
thanks

PR #1877, opened 2020-11-22
CSS versioning for cache busting after restart

DrAnax:
Puts a query string after main.min.css based on server start time. Used by browsers for CSS cache busting, so no need for hard reloads.

### DIFF
--- a/server/src/http.go
+++ b/server/src/http.go
@@ -26,6 +26,8 @@ type TemplateData struct {
 	Compiling   bool // True if we are currently recompiling the TypeScript client
 	WebpackPort int
 
+	CSSVersion int64 // for CSS Cache busting after server restart
+
 	// Profile
 	Name       string
 	NamesTitle string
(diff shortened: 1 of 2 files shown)


Zamiell:
that isn't a sufficient fix because i don't generally restart the server when updating the CSS

Zamiell:
i think that there is a much simpler fix, just have grunt name main.min.css in the same way that we name main.min.js

Zamiell:
i can set this up tomorrow or you can do another PR

Zamiell:
are there pros and cons to each?

DrAnax:
Idk, I asked to learn from others. From what I read there's no important advantage nowardays, certainly not given the current server's traffic.

As a personal preference, I'm not fond of changing filenames.

Zamiell:
ok, then submit a PR to use `?version=${version}` or whatever for both the JS and the CSS
(not server start time, but the git commit count)

Zamiell:
@DrAnax ok I worked on this a bit in this commit: https://github.com/Zamiell/hanabi-live/commit/08804301f2af8c3f2bf5bfd50857222c6dd77341

let me know if you want anything else done

DrAnax:
looks fine, thanks

Zamiell:
should i open an issue for removing the .hover class?

DrAnax:
no, it's a quickie, i'll do it asap

Στις Παρ, 27 Νοε 2020 στις 9:46 μ.μ., ο/η James <notifications@github.com>
έγραψε:

PR #1880, opened 2020-11-22
Center enter hypo button

timotree3:


client/src/game/ui/reactive/StateObserver.ts
```diff
@@ -307,6 +307,12 @@ const replayObservers: Subscriptions = [
   // game start
   subAfterInit((s) => s.replay.active, replayView.onActiveChanged),
 
+  // Enter Hypothetical button
+  subAfterInit(
+    (s) => s.finished && s.replay.active,
```
Zamiell, line 312: shouldn't this be `s.finished && s.replay.active && s.replay.shared === null`

timotree3: good catch. didn't test that case

Zamiell:
thanks

PR #1881, opened 2020-11-22
Lower End Hypo button during private replay

timotree3:


timotree3:
There is no chat button so it should go down next to Lobby

Zamiell:
ok, nice touch

PR #1886, opened 2020-11-25
remove reduntant code

DrAnax:
No need to add/remove row mouseover actions all the time; just once when the row is created.

Also the .hover class is redundant but I will remove it after the implementation of server CSS cache busting


Zamiell:
thanks

PR #1895, opened 2020-11-26
Impostor

vEnhance:
When the table leader issues `/impostor` the website will randomly tell one of the players they are an IMPOSTOR and the other players that they are a CREWMATE.

(Intended use case: players agree on a threshold N where the crewmates win if they either score at least N, or they successfully guess who the impostor is during review.)


Zamiell:
seems like server bloat but fine

PR #1900, opened 2020-11-27
Fix arrow rotation for own player in keldon mode

### LINKED ISSUE #1887: Turn indicator direction wrong when pointing down (Keldon mode)
Previous issue: #1844, addressed by #1845 

The new arrow direction looks good, except when pointing to the "down" player, when it points approximately northwest.

I haven't tested this with a game I'm playing, but I hopped in as a spectator for several games of different variants and numbers of players.

sample

timotree3:
Fixes #1887


Zamiell:
excellent, thank you very much

Issue #1864, opened 2020-11-21
add "Num. of Finesses Needed" to future eff tooltip


Zamiell:
"Num. of Finesses Needed" = the amount of 2-for-1 clues that you need to give before future efficiency reaches <= 1.00

Zamiell:
@vEnhance do you want to try implementing this?

vEnhance:
Sure. Is this just number of cards to get minus number of possible clues?

Zamiell:
"Num. of Finesses Needed" = the amount of 2-for-1 clues that you need to give before future efficiency reaches <= 1.00

so you have to project future eff by decrement cards gotten by 2 and decrement clues by 1
and then do that N times until future eff <= 1.00

timotree3:
Pretty sure both of your algorithms give the same answer

Zamiell:
i only described 1 algorithm

timotree3:
the two of you described one each

Zamiell:
oh, i see

DrAnax:
This is already implemented in the tooltip

Zamiell:
ok ill close then

DrAnax:
You haven't

Issue #1832, opened 2020-11-10
Feature Request - cards gotten modifier reads notes


glassbead0:
Certain notes on un-clued cards automatically increments the "cards gotten modifier" and thus the efficiency modifier. Which kinds of notes should trigger this is certainly up for debate.

Also decrement the cards gotten modifier for clued cards with a `kt` or `trash` note.

Maybe only be turned on if the Hyphenated Conventions are enabled. Everyone might not want this turned on.

Suggested notes that cause the increment:

`f`
`cm`
any identity note: (`b3`, `y2`)
identity noes separated by commas (or commas space?): (`r4, y4`, `g2, g3`, `3,4`)
maybe any partial identity note: (`4`, `b`)
...

Zamiell:
submit pull request

Zamiell:
cm wouldn't cause cards gotten to increase because you have to clue cm cards to get them to play

Zamiell:
@vEnhance should i close this now?

vEnhance:
Yep.

Zamiell:
ok thanks

PR #1902, opened 2020-11-27
Note eff mod

vEnhance:
Factors [f], [kt], [unclued] notes into efficiency calculation. Cards with [f][unclued] count as gotten. Any card with [kt] counts as not gotten regardless of the other notes.

Have tested, but I'm still paranoid I missed some cases (between specs, real replays, in-game replays, hypotheticals, etc.) so I would still appreciate some extra testing nonetheless.  :P


Zamiell:
in this neck of the woods we test in prod, and any bugs caused are your fault = )

vEnhance:
References #1832 and #1864.

vEnhance:
> in this neck of the woods we test in prod, and any bugs caused are your fault = )

xD sounds good

timotree3:
In the case that a card is `allPossibilitiesTrash` from clues and has no note on it, `notesEfficiencyModifier` will return -1?

client/src/game/ui/reactive/view/statsView.ts
```diff
@@ -52,6 +52,18 @@ export function onEfficiencyChanged(data: {
...
+  let cardsGottenByNotes: number | string;
+  if (!globals.state.finished && globals.state.playing) {
+    cardsGottenByNotes = 0;
+    for (const card of globals.deck) {
+      cardsGottenByNotes += card.notesEfficiencyModifier();
+    }
+    cardsGotten += cardsGottenByNotes;
+  } else {
+    cardsGottenByNotes = "-";
+  }
```
timotree3, line 65: Typically we don't access globals from code that is inside a view. I think this should be refactored to be a field in `StatsState` and be calculated from `statsReducer`. At that point we would also need a new action type for changing a note so that `statsReducer` will get called appropriately. Do you think you could do that? I could also do that as a follow-up.

We don't access globals from views because a view needs to be rerun every time that a value it depends on changes Imagine that `cardsGottenByNotes` is nonzero and the game is terminated. `onEfficiencyChanged` will not be called, but it needs to be called so that it recalculates the values with `globals.state.finished === true` such that it ignores notes.

vEnhance:
> In the case that a card is `allPossibilitiesTrash` from clues and has no note on it, `notesEfficiencyModifier` will return -1?

Mm good point. I was imagining that users would be typing `kt` notes but you're correct that we shouldn't assume this.

client/src/game/ui/reactive/view/statsView.ts
vEnhance: Makes sense, I'll give it a shot this weekend.

timotree3: Awesome. Nice PR btw. Thanks for your work

vEnhance:
Still WIP. As @timotree3 asked I moved the calculation into the stats reducer, but now the global variable referenced (namely `ourNotes`) is breaking the CI. I suspect I probably shouldn't be accessing globals in reducer anyways, and it would be better to have some sort of array of notes in the store.

Well, better than yesterday. Keep trying.

Zamiell:
its just failing on import version, which isn't your fault and is just a CI bug

Zamiell:
rebase your branch on master and the tests will pass

Zamiell:
probably

Zamiell:
> I think this should be refactored to be a field in StatsState and be calculated from statsReducer.

More discussion:

I explicitly did *not* do this for the global eff mod that I implemented, because of the following reason:
StatsReducer is a thing that re-computes a new state *if and only if* after a player takes an action.
In a game, a player can constantly be adding and removing an eff mod, so it didn't make sense at the time to have to constantly re-compute the entire game (e.g. N turns of state computation) as soon as someone changes the eff mod.

Now, in a game, a player can constantly be adding and removing notes in the same way, so I am unsure if we want the whole reducer process to be kicked off at the slightest note change.

Zamiell:
> I think this should be refactored to be a field in StatsState and be calculated from statsReducer.

I think this refactor would make more sense if the notes themselves were in state. Then you could leave the `state.replay.states` array as completely "real", and have modified eff in an object as follows:

```typescript
export default interface NotesState {
  readonly notes: Map<number, string>,
  readonly modifiedStatsStates: readonly StatsState[]
}
```

Zamiell:
If I haven't missed anything important in my previous 2 posts, then this project should be shelved until we have nuked globals.ourNotes + globals.allNotes and have all notes being handled by Redux State.

Once that is in place, then we can do this PR, which would add `readonly modifiedStatsStates: readonly StatsState[]` to the existing `NotesState` interface.

vEnhance:
oops, i just saw the new comments now (after finally sort of working around the issue)

it's too late for me to try writing more code (it kind of was already, to be honest). i agree though that moving `allNotes` and `ourNotes` into the state (rather than `global`) would basically make all the headache here go away. so i agree with the decision to do that first.

timotree3:
That makes sense, Zam. If we're going to put notes in the state in a separate PR, then I propose that we get this functionality merged now with a simpler fix in place for the bug I mentioned.

Let's go back to 97c42f7 and in order to fix the bug with `globals.state.finished` just change the `subVS` for `onEfficiencyChanged` to be a `subFullState` and include `state.playing` and `state.finished` in the selector.

Zamiell:
notes are in state now, which should make this PR more straightforward

Zamiell:
so now add the following to NotesState:

```
// We only track a modified version of the "present" StatsState and not any past turns for simplicity
readonly modifiedStatsState: readonly StatsState;
// Users can manually modify the efficiency with alt + right click
// This is not technically a note on a card, but we track it here
manualEfficiencyModifier: number;;
```

vEnhance:
Not done yet, but mostly there.

I did not implement a `modifiedStatsState` as I feel like it was easier to just add a `cardsGottenByClues` parameter. This ended up with the weird perk that the notes efficiency is remembered during the replay, i.e. if you rewind it will show the number at the time of the turn. Unless you F5 in which case of course the data goes away, because we don't keep track of when notes are added.

I plan to move efficiency modifier from globals in to `NotesState` but haven't done that yet.

vEnhance:
Should also mention that right now there's an optional argument to `gameStateReducer` for the `CardNote[]` array because I didn't want to go adding `null` into every other instance of game state reducer (e.g. in the tests, where we should just assume no notes on the cards, presumably).

I don't know if this project has a stance on whether optional arguments are permitted. If not permitted, then at some point I will go and add a bunch of `null` arguments all over the place :P

Zamiell:
> I did not implement a modifiedStatsState as I feel like it was easier to just add a cardsGottenByClues parameter.

sure

> This ended up with the weird perk that the notes efficiency is remembered during the replay, i.e. if you rewind it will show the number at the time of the turn.

i think the desired functionality is to only show modified eff in the present. so this would be a bug

> I plan to move the global manual efficiency modifier from globals in to NotesState but haven't done that yet.

ok, thanks

> I don't know if this project has a stance on whether optional arguments are permitted.

yes, they are permitted, use your best judgement

timotree3:
hmm. It is a little bit bad that note changes are gonna get encoded in the action log for replays now. is there any way to avoid that? maybe we can add a special case somewhere to ignore note actions?

Zamiell:
>  It is a little bit bad that note changes are gonna get encoded in the action log for replays now.

is that true? if so, that sounds like a bug

vEnhance:
timo are you referring to `state.replay.actions`? If so I don't think  that they get added in the way the code is written now? The `stateReducer` doesn't treat the `editNote` or `noteListPlayer` actions as game actions even though it recomputes the game state.

timotree3:
oh okay I believe you

Zamiell:
thanks evan for your work

PR #1904, opened 2020-11-27
remove redundant css and hidden button

DrAnax:
Removed unecessary hover class and the unecessary hidden button used for alt+j shortcut (used a class on the first table row)


Zamiell:
nice, thank you!

PR #1905, opened 2020-11-27
Enable unknown cards to be touched by a clue if all possibilities would be

timotree3:


Zamiell:
what does this do

timotree3:
If I start an in-game hypothetical and I have a clued 1, then I can clue it again it with 1 on other people's turns. Currently, it will not be touched by a 1 clue which is not ideal and can result in an empathy on the card where all the pips are missing because the card has both positive 1 and negative 1.

Zamiell:
okay, makes sense, thanks

Issue #1735, opened 2020-09-10
Game is confused about whose turn it is and no one can go


erinboyle:
Playing with detriments and rainbows. devolving alley agency. The last thing that happened was me bombing. I think what happened then is that the game auto-discarded my bomb card, and maybe one part of the logic therefore thinks I need to discard again since that's my detriment. But another part thinks it's not my turn so won't let me go. 

![image](https://user-images.githubusercontent.com/12789116/92677365-86fff300-f2d8-11ea-82aa-2edf3e3c7ba6.png)

padiwik:
This is with the Panicky (26) character

muyjohno:
+1 Had the same issue and came to the same conclusion (it occurred after a panicky misplay)

PR #1915, opened 2020-11-29
only increment discarded cards if not a strike

rkass:
Fixes https://github.com/Zamiell/hanabi-live/issues/1735

`cardsDiscardedThisTurn` is only used in [shouldEndTurnAfterDraw](https://github.com/Zamiell/hanabi-live/blob/master/client/src/game/rules/turn.ts#L7) when deciding if the panicky player's turn should continue. This change makes sure that this variable is not incremented after a misplay. 

**QA**
- tested that misplaying as panicky with < 5 clues advances turn
- tested that discarding, playing, cluing, and striking advance turn in normal cases

### DIFF
--- a/client/src/game/reducers/turnReducer.ts
+++ b/client/src/game/reducers/turnReducer.ts
@@ -43,7 +43,9 @@ function turnReducerFunction(
 
     case "discard": {
       turn.cardsPlayedOrDiscardedThisTurn += 1;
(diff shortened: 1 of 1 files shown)


Zamiell:
@timotree3 should review

timotree3:
Looks good. Thanks rkass

PR #1923, opened 2020-12-01
Note eff mod followup

vEnhance:
Refactor eff mod into notes state.

Also fix the quirk I mentioned about notes modifier being stored in memory during replays (now it always shows "-" instead).


Zamiell:
nice, looks good!

PR #1954, opened 2020-12-05
Fix contrarian endgame stuff

### LINKED ISSUE #1933: Contrarian only grants 1 turn after deck runs out
Game ID 360103

timotree3:
Fixes #1933 
Tested that game JSON and I was able to play p5 and finish the game.

Testing now if normal games finish properly

### DIFF
--- a/server/src/characters.go
+++ b/server/src/characters.go
@@ -685,6 +685,30 @@ func characterAdjustEndTurn(g *Game) {
 	}
 }
 
+func characterHasTakenLastTurn(g *Game) bool {
+	if g.EndTurn == -1 {
+		return false
+	}
+	originalPlayer := g.ActivePlayerIndex
+	activePlayer := g.ActivePlayerIndex
+	turnsInverted := g.TurnsInverted
+	for turn := g.Turn + 1; turn <= g.EndTurn; turn++ {
+		if turnsInverted {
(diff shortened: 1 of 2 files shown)


timotree3:
Actually, this shouldn't affect normal games since my new code only runs if there are detrimental characters. I guess then all there is to do is make sure the code looks right. i.e. no off-by-one errors in my for loop

Zamiell:
looks like it fails ci

Zamiell:
says that you need to delete an extra newline

timotree3:
> says that you need to delete an extra newline

it doesn't give a location

Zamiell:
its line 149, it should show you in vsode

Zamiell:
it does for me

Zamiell:
are you not using the linter? theres another blank line that the linter is complaining about that should be obvious if you just save the file and look in vscode

timotree3:
Yeah the linter doesn't seem to be working. Neither is find all references. Come to think of it, the only Go thing that's working is autoformat

Zamiell:
test to see if running `./server/lint_server.sh` works

Zamiell:
find all references doesn't work for me either, i think thats only a typescript feature

Zamiell:
thanks

PR #1955, opened 2020-12-05
Add number of finesses

vEnhance:
For #1864


Zamiell:
thank you, that looks easier than i thought it would be

PR #1964, opened 2020-12-06
Fix panicky Bottom-Deck Blind-Play

### LINKED ISSUE #1938: Player´s turn skipped
Game-ID: 360620

Player 1 plays the last card from the bottom of the deck. Then the turn of player 2 is kind of skipped. The game says that it is the turn of player 3, but the message "It is not your turn, so you cannot perform an action" appears when player 3 tries to play a card. Player 2 cannot do anything at all. Maybe an issue with detrimental characters (panicky, spiteful, miser)?

timotree3:
In the case of a bottom-deck blind-play, a `draw` action occurs that should not end the turn, so the condition of `cardsDiscardedThisTurn !== 1` is insufficient

Fixes #1938

### DIFF
--- a/client/src/game/rules/turn.ts
+++ b/client/src/game/rules/turn.ts
@@ -12,22 +12,17 @@ export function shouldEndTurnAfterDraw(
   variant: Variant,
 ): boolean {
   // Some "Detrimental Characters" are able to perform two actions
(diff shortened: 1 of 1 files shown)


Zamiell:
nice thanks

PR #1969, opened 2020-12-07
hypo: Fix morphing cards into their real identities

### LINKED ISSUE #1875: hypothetical: trying to morph a hidden drawn card into the card into actually is doesn't work

timotree3:
Fixes #1875


Zamiell:
looks good

timotree3:
By the way, this PR has the side-effect that now all drawn cards during in-game hypo will have the "blank" card back

Zamiell:
oh excellent

is there some other issue we can close then

PR #1983, opened 2020-12-07
Fix Escape in hypothetical

timotree3:


timotree3:
done testing

client/src/game/ui/keyboard.ts
```diff
@@ -105,7 +111,6 @@ function keydown(event: JQuery.KeyDownEvent) {
 
     // Escape = If in an in-game replay, exit back to the game
     replay.exit();
-    return;
```
Zamiell, line 108: should this be removed?

timotree3: good point

Zamiell:
thanks

PR #1989, opened 2020-12-08
Stats minor fix

vEnhance:
Two changes:

* If a card is known (e.g. we are Alice and see it in Bob's hand), then `allPossibilitiesTrash` will just use the known identity instead rather than the clued information. (Thus, trash cards essentially do not need `kt` notes anymore to get the efficiency correct.)

* Don't allow negative future efficiencies.

If a card is duplicated it will still count doubled until one of them plays, alas.


Zamiell:
nice thanks

PR #1996, opened 2020-12-09
align player buttons to log, fix timer overlap

Hyphen-ated:


Zamiell:
excellent, thank you very much

PR #2000, opened 2020-12-11
Css redesign

DrAnax:


client/src/lobby/keyboardInit.ts
```diff
@@ -30,8 +30,8 @@ export default function keyboardInit(): void {
       if (globals.currentScreen === Screen.Lobby) {
         $(".lobby-games-join-first-table-button").click();
       }
-    } else if (event.altKey && (event.key === "c" || event.key === "ç")) {
-      // Alt + c
+    } else if (event.altKey && (event.key === "n" || event.key === "ç")) {
```
padiwik, line 33: and the c cedille?

client/src/lobby/nav.ts
```diff
@@ -57,6 +57,43 @@ export function init(): void {
...
   });
 
+  // Games bottom screen toggle button
+  $("#lobby-toggle-show-tables").on("click", () => {
+    $("#lobby-toggle-show-tables").addClass("toggle-active");
+    $("#lobby-toggle-show-chat").removeClass("toggle-active");
+    $("#lobby-toggle-show-game-chat").removeClass("toggle-active");
+    $("#lobby-top-half").addClass("toggle-active");
+    $("#lobby-bottom-half").removeClass("toggle-active");
+    // restore online users and chat for small screens
```
Zamiell, line 67: capitalize comments, match the existing comment style

client/src/lobby/keyboardInit.ts
padiwik: ñ

DrAnax: Not sure what the second key is for; my guess was that it is mac specific. That's why I need a fix on line 33.

public/css/hanabi.css
```diff
@@ -833,72 +836,75 @@ input::-webkit-calendar-picker-indicator {
 }
 
 /*
-  Media edits
+  Media Responsive Utilities
```
Zamiell, line 839: good, thank you for documenting this

client/src/lobby/keyboardInit.ts
padiwik: Yeah it's mac specific https://www.webnots.com/option-or-alt-key-shortcuts-to-insert-symbols-in-mac-os-x/ so replace the ç with ˜ i guess

DrAnax:
done

Zamiell:
the breakpoint at 1280px seems off, can you confirm?
like, the new game button isn't even close to the "Hanab Live"

same thing with the breakpoint at 980px. it isn't as bad as the 1280px one, but it still needs to be adjusted

same with the one at 840px

Zamiell:
basically it should only break when the buttons start to overlap, or are at least close to overlapping

DrAnax:
Check again please

Zamiell:
better

but now the "Hanab Live" text disappears at 1065px

i think it should only disappear at something like 920px, e.g. when it gets close to overlapping

DrAnax:
Silly travis, I only touched a css file...

DrAnax:
All my cats agree, it should be purfect now

Zamiell:
yup, looking better

is it a bug that the rounded corners are removed in the mobile box?

DrAnax:
rounded corners of which element?

DrAnax:
It's an alpha feature. I guess it's meant to save screen space.

DrAnax:
alpha.css:3830

Zamiell:
i think you are wrong

just look at https://html5up.net/alpha

click on "elements", and then drag the window smaller, and you will see that even on mobile, it is supposed to look rounded

Zamiell:
anyways, the non-rounded corners look buggy, i think that the corners should always be rounded so that it looks consistent, if that is possible to do

DrAnax:
That's not what we use for background. We use the `box` class, an example of which can be found at https://html5up.net/alpha (it's the featured image container)

DrAnax:
Yes, it is

Zamiell:
this must have taken a long time to do, but its a huge upgrade for the website. very nice work!

PR #2005, opened 2020-12-11
Fix pregame css

DrAnax:


Zamiell:
?

DrAnax:
Didn't do the PR right, fixing it now

PR #2006, opened 2020-12-11
Fix bug, put chat boxes side by side if possible

DrAnax:
Fixed annoying pregame scrollbar. Put chatboxes side by side if screen is wide enough


Zamiell:
good, thanks

PR #2007, opened 2020-12-12
Fix pregame

DrAnax:


Zamiell:
thanks

DrAnax:
Check if it is ok now, if not pm

Zamiell:
yeah looks better!

PR #2028, opened 2021-01-07
Implement negative identity notes

### LINKED ISSUE #1801: Should be able to make notes about what a card isn't and it gets crossed off on the card
something like !2g and makes the card visually not green 2

timotree3:
Resolve #1801

`!b,g` means not blue and not green.

### DIFF
--- a/client/src/game/reducers/noteIdentity.ts
+++ b/client/src/game/reducers/noteIdentity.ts
@@ -47,6 +47,28 @@ export function parseIdentity(variant: Variant, keyword: string): CardIdentity {
   return { suitIndex, rank };
 }
 
+function possibilitiesComplement(
+  variant: Variant,
+  possibilities: Array<[number, number]>,
+): Array<[number, number]> {
+  const complement: Array<[number, number]> = [];
+  for (let suitIndex = 0; suitIndex < variant.suits.length; suitIndex++) {
+    for (const rank of variant.ranks) {
+      // Ensure that this possibility is not present in the source list
+      if (
+        !possibilities.some(
+          ([negatedSuitIndex, negatedRank]) =>
+            negatedSuitIndex === suitIndex && negatedRank === rank,
+        )
+      ) {
+        complement.push([suitIndex, rank]);
+      }
+    }
+  }
+
+  return complement;
+}
+
 // Examines a single square bracket-enclosed part of a note (i.e. keyword) and returns the set of
 // card possibilities that it declares
 //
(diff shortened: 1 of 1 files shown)


Zamiell:
thank you

Zamiell:
just double checking, does this pr implement my preferred syntax, as described in the other thread?

timotree3:
With this PR `!g2,b2` means any card besides (green 2 or blue 2) and `!g2,!b2` would fail to parse

Zamiell:
you do agree that my syntax is more intuitive / better?

timotree3:
No. I would prefer that `gb` = `g,b` and `!gb` means not (green or blue). (And therefore `!g,b` means that as well) Generally expressing `(anything besides X) OR (Y)` is not useful since it's the same as `(anything besides X)`. It happens that for my argument above I picked the one case where it is kind of useful: where X is an entire suit and Y is an entire rank

Zamiell:
i think parsing `gb` as `green or blue` is problematic because users have historically used two letters to abbreviate a single suit (and will continue to do in the future), e.g. dark brown as `db`

from this it follows that a separator must be used, and the conventional separator is commas

timotree3:
> i think parsing `gb` as `green or blue` is problematic because users have historically used two letters to abbreviate a single suit (and will continue to do in the future), e.g. dark brown as `db`

I don't think that's much of a problem. It seems unlikely that both letters will correspond to a valid suit shorthand, and I really desire for the website to support the note shorthand that I prefer where `gb3` means `green 3, blue 3`

Zamiell:
the problem is that supporting `!gb` means that the `!g, b` case is extremely unintuitive for normal users

Zamiell:
how do you propose that a normal user expresses `!g, !b, 3`

timotree3:
[3] [!g,b]

Zamiell:
i think that most users of the website use the comma notation, so it seems weird for the "squishing" deviators to be forcing extraneous brackets on normal users

to resolve this conflict, can/should we parse commas differently from letters that are squished together?
e.g.
`!g, !b` === `!gb`
`!g, !b` === `!gb[3]` (feel free to revise this one)

padiwik:
I think the best syntax is one where squishing is allowed everywhere except between a rank and a suit. If there is a comma between a rank and a suit, this starts a new rule. Similarly, a new exclamation mark starts a new rule (Isn't this the most uncontroversial rule? I'm not sure why this wasn't implemented)

Then:
`!gb` = `!g,b` = `!g,!b` (not a card matched by `gb` = not a card matched by `g` NOR a card matched by `b`)
`!gb3` = `!g3,!b3` (not a card matched by `gb3`)
`!gb,3` = `!g,b,3` = `[3], !g, !b` (not a card matched by `gb` AND is a card matched by `3`)
`!g3,b3` = `!g3, !b3` (not a card matched by `g3` NOR a card matched by `b3`)
`!r,m2` = `!r, !m2` (not a card matched by `r` NOR a card matched by `m2`)
etc.

I believe this satisfies everyone's comments so far. It makes sense because something like `b,!g` is nonsense. Therefore, `!g,b` can be shorthand for `!g,!b`.

"Not green and not a 2" would require the long-form syntax of `!g,!2` but that's fine because this situation is very rare.

marcmagus:
I may also not be a "normal" user, but fwiw my intuition is that `gb` should be available and mean (g||b), `!gb` should be available and mean !(g||b), and that `!g!b` should be available as a synonym for the same.

If I wanted to express (3 && !(g||b)), I would try in order something like "!gb,3", "3!gb", "!gb 3", "{!g,!b,3}". I would eventually try "[!gb][3]" because I have read the site docs, but I still have to remind myself to use [] as a separator here.

padiwik:
Nice, my proposal matches perfectly with what marcmagus intuitively tries as a user. But it seems `gb` = `(g||b)` is not something that exists yet, unfortunate.

PR #2029, opened 2021-01-07
Confirm `/suggest` less often

### LINKED ISSUE #2025: change suggest message from replay to shared replay
Issue opened by Discord user: Valetta6789#5653

timotree3:
Fixes #2025 assuming I understand the issue correctly.

### DIFF
--- a/client/src/game/ui/HanabiUI.ts
+++ b/client/src/game/ui/HanabiUI.ts
@@ -57,6 +57,8 @@ export default class HanabiUI {
 
   // eslint-disable-next-line class-methods-use-this
   suggestTurn(who: string, segment: number): void {
+    // We minus one to account for the fact that turns are presented to the user starting from 1
+    const internalSegment = segment - 1;
     if (
       globals.state.finished &&
       globals.state.replay.shared !== null &&
@@ -64,10 +66,11 @@ export default class HanabiUI {
       globals.state.replay.hypothetical === null
     ) {
       if (
+        internalSegment === globals.state.replay.shared.segment ||
+        who === globals.metadata.ourUsername ||
         window.confirm(`${who} suggests that we go to turn ${segment}. Agree?`)
       ) {
-        // We minus one to account for the fact that turns are presented to the user starting from 1
-        replay.goToSegment(segment - 1);
+        replay.goToSegment(internalSegment);
       }
     }
   }
--- a/client/src/game/ui/HanabiUI.ts
+++ b/client/src/game/ui/HanabiUI.ts
@@ -2,6 +2,7 @@
 // It is re-created every time when going into a new game
 // (and destroyed when going to the lobby)
 
+import { addSelf } from "../../chat";
 import { Globals as LobbyGlobals } from "../../globals";
 import { GameExports } from "../main";
 import * as cursor from "./cursor";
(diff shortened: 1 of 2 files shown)


Zamiell:
furthermore, ensure:

- typing /suggest when not in a shared replay should give an error message
- typing /suggest when you are the shared replay leader should give an error message

timotree3:
> typing /suggest when you are the shared replay leader should give an error message

nah. that can just take you to that turn

Zamiell:
disagree.
the only reason you would be typing that is if you did not already know that you were the shared replay leader, thus there should be feedback informing that explicitly

timotree3:
or if you don't know that you can click on the turn number

Zamiell:
furthermore, suggesting the current turn should also result in an error message

Zamiell:
what is the point of the who argument in this function, can it be removed

timotree3:
I don't want to give error messages that disrupt the flow of using the site when nothing is wrong. A player typed `/suggest` at the same time as the replay leader took us to that turn? Great!

timotree3:
> what is the point of the who argument in this function, can it be removed


You don't see both uses of the argument in the function body??

timotree3:
> typing /suggest when not in a shared replay should give an error message

This already errors as desired.

Zamiell:
it looks like the code you wrote is suggesting a turn to yourself

Zamiell:
who === globals.metadata.ourUsername

timotree3:
Do you realize that this is the part of the client that receives `/suggest` from any other player as the shared replay leader?

Zamiell:
if it is from another player, then the window.prompt line will never be executed

Zamiell:
> A player typed /suggest at the same time as the replay leader took us to that turn? Great!

fine, but do the other error message

timotree3:
> if it is from another player, then the window.prompt line will never be executed

????? That's not true. I have no idea why you're saying these things. Maybe read the code again?

Zamiell:
you're right, i was misreading the code

i maintain that the `who === globals.metadata.ourUsername` line is a bug and should be removed, because we should never be suggesting turns to ourselves, and should instead throw an error message

timotree3:
Now it gives an informational message in that case.

timotree3:
Screenshot from 2021-01-07 15-27-34

client/src/game/ui/HanabiUI.ts
```diff
@@ -56,18 +57,28 @@ export default class HanabiUI {
...
+        internalSegment === globals.state.replay.shared.segment ||
         window.confirm(`${who} suggests that we go to turn ${segment}. Agree?`)
       ) {
-        // We minus one to account for the fact that turns are presented to the user starting from 1
-        replay.goToSegment(segment - 1);
+        replay.goToSegment(internalSegment);
+      }
+      if (leaderSuggested) {
+        addSelf(
+          "As shared replay leader, consider clicking on the turn number to change turns rather than using /suggest",
```
Zamiell, line 79: your sentence does not have any punctuation. try this instead:

"You are the shared replay leader, so you can simply click on the turn number instead of using the <code>/suggest</code> command."

Zamiell: im not sure if the `<code>` part will work properly, you'll have to try it out

timotree3:
Screenshot from 2021-01-07 15-34-13

> im not sure if the <code> part will work properly, you'll have to try it out

Seems to be working.

Zamiell:
nice, thanks for your work

PR #2033, opened 2021-01-07
Implement delay before being able to start table

### LINKED ISSUE #1976: disable start button for 1s after someone joins/leaves

timotree3:
Fixes #1976

The UI could use improvement but what do you think about this approach?

### DIFF
--- a/server/src/command_table_start.go
+++ b/server/src/command_table_start.go
@@ -66,8 +66,9 @@ func commandTableStart(ctx context.Context, s *Session, d *CommandData) {
 			}
 		}
 	}
-
-	if now.Sub(t.DatetimeLastJoined) < time.Millisecond*1500 || now.Sub(t.DatetimeLastLeft) < time.Millisecond*1500 {
+	recencyThreshold := 1500 * time.Millisecond
+	playersRecentlyChanged := now.Sub(t.DatetimeLastJoined) < recencyThreshold || now.Sub(t.DatetimeLastLeft) < recencyThreshold
+	if playersRecentlyChanged {
 		s.Warning("You must wait 1.5 seconds after a player joins or leaves before starting the game.")
 		return
 	}
--- a/server/src/command_table_start.go
+++ b/server/src/command_table_start.go
@@ -67,8 +67,8 @@ func commandTableStart(ctx context.Context, s *Session, d *CommandData) {
 		}
 	}
 
-	if now.Sub(t.DatetimeLastJoined) < time.Second*2 || now.Sub(t.DatetimeLastLeft) < time.Second*2 {
-		s.Warning("You must wait 2 seconds after a player joins or leaves before starting the game.")
+	if now.Sub(t.DatetimeLastJoined) < time.Millisecond*1500 || now.Sub(t.DatetimeLastLeft) < time.Millisecond*1500 {
+		s.Warning("You must wait 1.5 seconds after a player joins or leaves before starting the game.")
 		return
 	}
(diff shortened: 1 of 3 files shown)


server/src/command_table_start.go
```diff
@@ -65,6 +67,11 @@ func commandTableStart(ctx context.Context, s *Session, d *CommandData) {
 		}
 	}
 
+	if now.Sub(t.DatetimeLastJoined) < time.Second*2 || now.Sub(t.DatetimeLastLeft) < time.Second*2 {
+		s.Warning("Cannot start game immediately after a player joins or leaves. Try again in 2 seconds")
```
Zamiell, line 71: make this match the style of the existing error messages

Zamiell:
suggestion: can we just nuke datetimelastjoined and datetimelastleft and simply use datetimelastaction instead?
(remember to reinitialize it to X during game start)

timotree3:
> suggestion: can we just nuke datetimelastjoined and datetimelastleft and simply use datetimelastaction instead?
> (remember to reinitialize it to X during game start)

I don't think any action should prevent game start. e.g. changing variant shouldn't. Just joining and leaving

Zamiell:
sure, then simply dont have /changevariant affect datetimelastaction

timotree3:
lastaction is currently only used for killing idle tables. I think that concern is unrelated to this PR's and does care about changing variant

server/src/command_table_start.go
```diff
@@ -65,6 +67,11 @@ func commandTableStart(ctx context.Context, s *Session, d *CommandData) {
 		}
 	}
 
+	if now.Sub(t.DatetimeLastJoined) < time.Millisecond*1500 || now.Sub(t.DatetimeLastLeft) < time.Millisecond*1500 {
```
Zamiell, line 70: try this:

```go
recentTimeThreshold := 1.5 * time.Second
playersRecentlyChanged := now.Sub(t.DatetimeLastJoined) < recentTimeThreshold  || now.Sub(t.DatetimeLastLeft) < recentTimeThreshold 
if playersRecentlyChanged  {
```

timotree3: I doubt you can multiply a duration by a float

Zamiell: if you can't multiply a duration by a float, then do:

```go
recentTimeThreshold := 1.5 * time.Millisecond
playersRecentlyChanged := now.Sub(t.DatetimeLastJoined) < recentTimeThreshold || now.Sub(t.DatetimeLastLeft) < recentTimeThreshold 
if playersRecentlyChanged {
```

timotree3:
I was gonna implement graying out the start game button momentarily after a player joins but I don't think that's worthwhile. Especially since in many of these cases the client will learn too late that someone joined. I believe this is ready to merge and we'll get feedback from users as to tweaking to 1.5 seconds constant

Zamiell:
i do think that graying out the start game button would be appropriate.
the avg amount of milliseconds from server to client is 100ms, which is an order of magnitude less than 1.5 seconds

timotree3:
Alright I'm gonna go for a different approach then.

Zamiell:
also i should probably mention that we should put a freeze on any server PRs until channels is merged

timotree3:
I don't think that's necessary. Just merge my changes into your channels branch

Zamiell:
have you seen the code? its like completely different

timotree3:
I haven't

Issue #2053, opened 2021-01-19
Feature Request: Set maximum player count on table creation


eskolnik:
In the current implementation, any public table will allow 6 players to join, which can lead to either unenjoyable games (as many folks prefer 3-5p to 6p) or the group needing to coordinate to split into separate lobbies. The simplest solution to this would be to cap lobbies at 5, but for the sake of people who do enjoy 6p games, it would be nice to leave those available. To that end, I propose that we have newly created tables default to a 5-player cap, but provide an option in the table creation UI for the host to set an arbitrary player cap (2-6). 

Example UI
![image](https://user-images.githubusercontent.com/1791642/105071070-1e790d80-5a52-11eb-9483-9063b2d04fa6.png)

DrAnax:
Fixed via #2314

PR #2055, opened 2021-01-20
Maximum player option

eskolnik:
This is an attempt to address the 6-player issue raised in #2053 
This PR allows the creator of a table to set a player limit, defaulting to 5. The current / max player count is displayed on each unstarted table in the lobby.

I opted to put the player count option in the main game creation modal, but if we want to add a little friction (and thereby reduce the number of 6p lobbies even further), we could nest it under "Extra Options." 

hanabUI_max_players_create
hanabUI_max_players_lobby
hanabUI_max_players_warning

### DIFF
--- a/client/src/lobby/createGame.ts
+++ b/client/src/lobby/createGame.ts
@@ -257,7 +257,7 @@ function submit() {
   }
   localStorage.setItem("createTablePassword", password);
 
-  // Max players must be a number between 2 and 6, default to 5 if validation fails
+  // Max players must be a number between 2 and 6. Default to 5 if validation fails
   let maxPlayers = Number($("#createTableMaxPlayers").val());
   if (maxPlayers < 2 || maxPlayers > 6) {
     maxPlayers = 5;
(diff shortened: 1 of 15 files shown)


client/src/lobby/createGame.ts
```diff
@@ -257,6 +257,12 @@ function submit() {
   }
   localStorage.setItem("createTablePassword", password);
 
+  // Max players must be a number between 2 and 6, default to 5 if validation fails
```
timotree3, line 260:
```suggestion
  // Max players must be a number between 2 and 6. Default to 5 if validation fails
```

timotree3 (approved):
Thanks! The code looks great. We'll see if Zamiel wants to merge the feature

Zamiell:
seems fine

Zamiell:
i can't merge any server changes until the channels branch is merged though

eskolnik:
fair enough, I figure just having this PR to point to should be sufficient to (hopefully) resolve the 6p discussion, whether it gets merged now or weeks from now.

Zamiell:
also a semi colon is more grammatically correct than a comma

Zamiell:
looks like this is fixed in #2314

PR #2071, opened 2021-01-30
/ping and /unping

Pimak:
Added the commands /ping and /unping
/ping add the role Ping
/unping remove the role

The Ping role is hard-coded

This is untested


Zamiell:
if its untested, then test it

padiwik:
> The Ping role is hard-coded

have a look at .env for how it should be done

PR #2107, opened 2021-02-26
Fix lag in throw it in a hole 

squidgetx:
Right now when a new card is added to the stack `doLayout` is called, which attempts to animate every card on the stack. This causes insane lag when the play stack is large, which is especially notable in Throw It In A Hole where every card is attached to the same stack object.

I don't see why we can't just animate the most recently played card. 

I tested no variant, TIH, spectator, and replay and everything works fine (and with less lag)

### DIFF
--- a/client/src/game/ui/PlayStack.ts
+++ b/client/src/game/ui/PlayStack.ts
@@ -19,47 +19,47 @@ export default class PlayStack extends Konva.Group {
   doLayout(): void {
     const lh = this.height();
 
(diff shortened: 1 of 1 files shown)


Zamiell:
your build is (obviously) failing the linter, make it pass the linter, this is covered in the install.md file

squidgetx:
linter passed

Zamiell:
smart fix, well done

Zamiell:
there's plenty to do around these parts if you interested in fixing more bugs

PR #2109, opened 2021-02-27
Allow marking cards as critical with notes

squidgetx:
This would be useful for throw it in a hole when Alice misplays a card and Bob wants to mark the other copy in Cathy's hand as critical.

I wrote it such that it only works on other people's cards.

### DIFF
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -813,11 +813,14 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   }
 
   private shouldSetCritical(critical: boolean) {
+    // use the critical note if it's on another player's card
     return (
-      critical &&
-      !cardRules.isPlayed(this.state) &&
-      !cardRules.isDiscarded(this.state) &&
-      !globals.lobby.settings.realLifeMode
+      (this.note.critical &&
+        this.state.location !== globals.metadata.ourPlayerIndex) ||
+      (critical &&
+        !cardRules.isPlayed(this.state) &&
+        !cardRules.isDiscarded(this.state) &&
+        !globals.lobby.settings.realLifeMode)
     );
   }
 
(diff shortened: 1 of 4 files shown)


Zamiell:
why does it only work on other people's cards?

squidgetx:
Hm, originally I added that logic so that you couldn't mark an unknown card as critical (since that wouldn't make much sense, and is visually ugly). 

But it probably could be useful for example: if you have a card that you've marked as b4 and the other b4 misplayed, you would want to add the critical marker to your own b4 "[b4] [critical]."

timotree3:
> But it probably could be useful for example: if you have a card that you've marked as b4 and the other b4 misplayed, you would want to add the critical marker to your own b4 "[b4] [critical]."

It seems like if card has a note on it as though it's fully known blue 4, and blue 4 is critical, then we should display the critical marker automatically.

Zamiell:
it just seems like the most consistent thing to let users morph whatever cards they want.
e.g. kt notes can be on any card, the client doesn't care

Zamiell:
> It seems like if card has a note on it as though it's fully known blue 4, and blue 4 is critical, then we should display the critical marker automatically.

i agree, if it doesn't do this already, then that seems like a bug

squidgetx:
Happy to change it to be more flexible. 

In non Throw It In A Hole variants the client does mark all known critical cards automatically. The client correctly does not label cards that are critical because the other copy was misplayed, since you can't "see" the other copy.

Zamiell:
oh, ok

client/src/game/ui/HanabiCard.ts
```diff
@@ -813,10 +813,9 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   }
 
   private shouldSetCritical(critical: boolean) {
-    // use the critical note if it's on another player's card
+    // use the critical note if present
```
Zamiell, line 816: all comments must be capitalized

Zamiell:
whats the status on this?

Issue #2098, opened 2021-02-18
Impossible Notes


Smandon1:
When playing 'Throw It in a Hole', I often put notes on the play piles to keep track of what I think happened.  Sometimes during the game I am unable to enter notes for a play pile, typically the green play pile (I'm not sure if any other piles are affected).  Refreshing fixes the issue.

To reproduce:
1) Start a 'Throw It in a Hole' game.
2) Play a card.
3) Right click on the green pile, in a (oftentimes) feeble attempt to leave a note.  It will not pop up with the note input box.

This doesn't always happen for all players after playing a card, but it does seem to be happen > 50% of the time, from limited testing.

DrAnax:
Closed via #2110

Zamiell:
ok thanks

Issue #2097, opened 2021-02-18
Morphed Play Piles In Replay


Smandon1:
When playing 'Throw It in a Hole', I often put notes on the play piles to keep track of what I think happened.  When the game ends, those piles continue to show the last morphed identity.  If I go back into my history and view a game then the piles look fine.

To reproduce:
1) Start a 'Throw It in a Hole' game.
2) Put a '1' note on the red play pile.
3) End the game

In the immediate replay created, the play pile will still show a red 1 when it should be empty.

DrAnax:
According to #2110 this has been fixed and it should be closed

Zamiell:
ok

PR #2110, opened 2021-02-28
Fixing notes issues (#2097 and #2098) with throw it in a hole.

Smandon1:
#2097 : Unmorphing play stacks at game end by adding a new observer on the "finished" property.
#2098 : Allowing notes on a middle play stack after someone plays a card by creating the invisible "hole" play stack before the regular play stacks, so the "hole" will be behind the middle play stack as opposed to in front of it (so it won't block notes anymore).


Zamiell:
nice thank you!

Issue #1225, opened 2020-06-12
Notes visual effects


TheDaniMan:
I like the visual effects from notes for things like r4/f/cm/fixme/trash; can these visual effects be made to remain even if more notes are added?
For example, I'd like to be able to write cm | probably r3 and still have a white border to remind me to not discard, or f | g1 to have a blue border but also note down that I'm trying to play g1.

robertos:
Use brackets :) "[cm] something else" will keep the border

robertos:
cm | probably r3 -> [cm] | probably r3
  f | g1 -> [f] | [g1]

padiwik:
Which involves retyping the note, which is annoying. I understand zam wants piping something to a note to remove the border, but imo it's more likely we want to add information to a cm/finesse than turn off the cm/finesse

robertos:
i see | as a "replace"/update symbol. what's left of it is just for history. Maybe a different symbol like "+"? 
e.g., "cm + r1" = add all visual effects from cm and r1
and "cm | f + r1" = only f and r1 effects, cm is history

Zamiell:
> Which involves retyping the note, which is annoying.

no, it doesn't. just put brackets around the existing note

PR #2116, opened 2021-03-04
Appending notes uses bracket syntax

### LINKED ISSUE #2108: Improve the cm, and f notes
I think i would be helpfull if the right click + ctrl or ALt writes [cm] [f] so we can just add up our notes without erasing the border .

newduke:
fixes #2108

### DIFF
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -1122,16 +1122,20 @@ export default class HanabiCard extends Konva.Group implements NodeWithTooltip {
   }
 
   appendNote(note: string): void {
-    // By default, set the note directly on the card
-    let newNote = note;
-
-    // If we had an existing note, append the new note to the end using pipe notation
     const existingNote = globals.state.notes.ourNotes[this.state.order].text;
+    let noteText = "";
     if (existingNote !== undefined && existingNote !== "") {
-      newNote = `${existingNote} | ${note}`;
+      const lastPipe = existingNote.lastIndexOf("|");
+      let lastNote = existingNote.slice(lastPipe + 1);
+      if (lastNote.trim()[0] !== "[") {
+        const bracketedNote = `[${lastNote}]`;
+        lastNote = bracketedNote;
+      }
+      // If we had an existing note, append the new note to the end using new brackets
+      noteText =
+        (lastPipe === -1 ? "" : existingNote.slice(0, lastPipe)) + lastNote;
     }
-
-    this.setNote(newNote);
+    this.setNote(`${noteText}[${note}]`);
   }
 
   checkSpecialNote(): void {


Zamiell:
no, we dont want to change this functionality

Zamiell:
can you submit a pr instead that only looks for square brackets in the rightmost pipe section?
that fixes the problem without inserting a bunch of stupid brackets everywhere

Zamiell:
i should of just programmed it that way from the beginning but i was a stupid person

newduke:
Um. Why you close?

Zamiell:
can you submit a pr instead that changes the note-logic to only search for square brackets in the rightmost pipe section?
that fixes the problem without inserting a bunch of stupid brackets everywhere

padiwik:
That's a nice idea but does not fix the problem described in the linked issue.

newduke:
@padiwik Why not? Please elaborate.

padiwik:
@newduke If you write `cm | probably r3` or even `cm probably r3` then neither of these would work. In fact Zam's proposal breaks `[cm] | probably r3` (currently suggested in #1225) which might be undesirable.

newduke:
don't you want `this is a note|[r2][cm]`?  The whole point is to not lose previous notes.

padiwik:
The issue is when you need to append additional notes after doing ctrl + right click or whatever

newduke:
Please refute `this is a note|[r2][cm]`

Zamiell:
```
cm | probably r3
```
that's a wrong note, someone would instead write
```
[cm] probably r3
```

padiwik:
> Please refute `this is a note|[r2][cm]`

Are you suggesting this as an alternative syntax? If you have a `cm` card and then need to mark it as `probably r3` you would not put that information before the `cm`. It should be like Zam pointed out.

Zamiell:
and then, that would be piped when new actual information is received
e.g.
```
[cm] probably r3 | [cm] [r3]
```

Zamiell:
> The issue is when you need to append additional notes after doing ctrl + right click or whatever

appending notes adds a pipe, therefore i dont see what you are getting at

padiwik:
By appending notes I mean adding info like "probably r3", which I understand is different from some other contexts of appending where you do actually need a pipe.

Zamiell:
im fine with that, people will learn the new syntax and adjust accordingly

padiwik:
Sure, fine, but they still would need to manually add brackets to the note when adding info like `probably r3`. If the keyboard shortcuts automatically add brackets and brackets are only parsed on the right-most piped section, everything works well.

Zamiell:
> they still would need to manually add brackets to the note when adding info like probably r3

this is how it currently is, so i dont follow

padiwik:
> this is how it currently is, so i dont follow

This is the entire issue multiple people are complaining about, and which this PR is trying to fix.

Zamiell:
this pr does not fix `probably r3`

Zamiell:
the behavior for that note is unchanged

padiwik:
With the PR: You start off with `[cm]` and you only need to add the text ` probably r3`.
Without the PR: You start off with `cm` and need to rewrite the entire note as `[cm] probably r3` which is silly.

padiwik:
I should probably also share my opinions on what you call "needless" brackets. I believe `[f]` looks nicer than `f`, especially because it makes clear that this is special syntax that the game interprets a certain way. There's no downside to changing the keyboard shortcut to doing that, especially if we also add the "brackets only read in right-most section" thing. At any rate, a notion of note aesthetics should not prevent functionality that many people want.

newduke:
So... reopen please?

Zamiell:
ok i'll accept this but we also need the other pr too

Zamiell:
i disagre that `[f]` "looks nicer" than `f`, i think it looks cluttered and stupid, but the other thing you said was convincing

padiwik:
This is basically irrelevant but `f` just feels too naked or bare to me. For other notes like `[cm]` I can understand why it seems cluttered to you.

Zamiell:
i love being naked

Zamiell:
and bare

Zamiell:
yes please can you do this pr https://github.com/Zamiell/hanabi-live/pull/2116#issuecomment-790909266

newduke:
> yes please can you do this pr [#2116 (comment)](https://github.com/Zamiell/hanabi-live/pull/2116#issuecomment-790909266)

That is precisely what this PR does, no?

padiwik:
If you have a note like `[foo] | [bar]` then only bar should be read.

padiwik:
This is the PR Zam proposes. Presumably to allow appending notes such as `[cm]` -> `[cm] | not cm` and then the card is no longer cm.

PR #2117, opened 2021-03-04
Nonleaders can suggest variants using dropdown

newduke:
No issue. I just found it annoying that suggesting variants as a nonleader I would have to remember variant names exactly but there is already a perfectly good UI for that.

### DIFF
--- a/client/src/chat.ts
+++ b/client/src/chat.ts
@@ -208,6 +208,7 @@ function send(room: string, element: JQuery<HTMLElement>) {
     throw new Error("The value of the element is not a string.");
   }
   msg = msg.trim();
+
   // Validate that they are accidentally broadcasting a private message reply
   if (msg.startsWith("/r ")) {
     modals.warningShow(
--- a/client/src/chat.ts
+++ b/client/src/chat.ts
@@ -222,9 +222,9 @@ function send(room: string, element: JQuery<HTMLElement>) {
   sendText(room, msg);
 }
 
-export function sendText(room: string, _msg: string): void {
+export function sendText(room: string, msgRaw: string): void {
   // Validate that they did not send an empty message
-  if (_msg === "") {
+  if (msgRaw === "") {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


client/src/chat.ts
```diff
@@ -220,8 +219,12 @@ function send(room: string, element: JQuery<HTMLElement>) {
   // Clear the chat box
   element.val("");
 
+  sendText(room, msg);
+}
+
+export function sendText(room: string, _msg: string): void {
```
Zamiell, line 225: get rid of the underscore

newduke: Lint complains about reassigning parameters. Prefer to rename to msgConst or something?

Zamiell: if you have to reassign a parameter, then make a new variable that has a name describing how it was modified

newduke: I dislike `msgWithEmojis`.

Zamiell: maybe formattedMsg?

newduke: Sure, but the only format changed is that last emoji... it really is `msg` so I think `msgRaw` is smart naming

newduke: I guess it's less pretty as an export

Zamiell: i dont feel strongly either way

client/src/chat.ts
```diff
@@ -208,7 +208,6 @@ function send(room: string, element: JQuery<HTMLElement>) {
     throw new Error("The value of the element is not a string.");
   }
   msg = msg.trim();
-
```
Zamiell, line 211: why remove this

client/src/lobby/nav.ts
```diff
@@ -148,12 +149,16 @@ export function init(): void {
         if (VARIANTS.get(variantName) === undefined) {
           return;
         }
-        globals.conn!.send("tableSetVariant", {
-          tableID: globals.tableID,
-          options: {
-            variantName,
-          },
-        });
+        if (globals.game!.owner === globals.userID) {
```
Zamiell, line 152: probably better written as `game?` right?

newduke: Fine by me.

Zamiell:
thx ill test this out later

PR #2128, opened 2021-03-11
Search through chat history

newduke:
Start typing `Black`. Hit `up`.  Have it autocomplete everything you've said which started `Black...`

Hit `<enter>`.  Now that chat is pushed to the beginning of chat history so you don't have to scroll back to find it.

### DIFF
--- a/client/src/chat.ts
+++ b/client/src/chat.ts
@@ -332,7 +332,7 @@ function historyMatchNext(current: string, increment: number): string | null {
     typedChatHistoryPrefix = current;
   }
   const oldIndex = typedChatHistoryIndex;
-  while (1) {
+  do {
     typedChatHistoryIndex += increment;
     // Stay within bounds of history
     if (typedChatHistoryIndex >= typedChatHistory.length) {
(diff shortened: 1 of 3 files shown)


client/src/chat.ts
```diff
@@ -13,6 +13,7 @@ import * as modals from "./modals";
 import ChatMessage from "./types/ChatMessage";
 
 // Constants
+const TYPED_HISTORY_MAX_LENGTH = 250;
```
Zamiell, line 16: i guess this shouldn't be capitalized because it isn't exported (as per Airbnb)

newduke: But it is a "constant" (not just a const).  I'm going to move it into a constants file :)

client/src/chat.ts
```diff
@@ -305,49 +310,59 @@ function keydown(this: HTMLElement, event: JQuery.Event) {
...
-    typedChatHistoryIndex = finalIndex;
-    return;
+    if (increment < 0) {
+      return null;
+    }
+    typedChatHistoryIndex = -1;
+    typedChatHistoryPrefix = current;
+  }
+  const oldIndex = typedChatHistoryIndex;
+  while (1) {
```
Zamiell, line 335: this should be rewritten without a constant condition, as per airbnb

newduke: `for (;;)` better?

newduke: I cannot find that section of airbnb guide.

Zamiell: https://eslint.org/docs/rules/no-constant-condition

Zamiell: like instead of doing an infinite loop, the code should be restructured such that the while loop contains the actual condition that the loop should be broken on

newduke: Done. It is better w/ real loop condition.

Zamiell: ><

PR #2129, opened 2021-03-11
Improved appended notes

newduke:
Follows convention from chat w/ @Zamiell

### DIFF
--- a/client/src/game/reducers/noteIdentity.ts
+++ b/client/src/game/reducers/noteIdentity.ts
@@ -178,7 +178,8 @@ export function getPossibilitiesFromKeywords(
 ): Array<[number, number]> {
   let possibilities: Array<[number, number]> = [];
 
-  for (const keyword of keywords) {
+  // Empty keyword list returns all possibilities
+  for (const keyword of keywords.length ? keywords : [""]) {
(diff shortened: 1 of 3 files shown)


Zamiell:
horray

client/src/game/reducers/noteIdentity.ts
```diff
@@ -178,7 +178,8 @@ export function getPossibilitiesFromKeywords(
 ): Array<[number, number]> {
   let possibilities: Array<[number, number]> = [];
 
-  for (const keyword of keywords) {
+  // Empty keyword list returns all possibilities
+  for (const keyword of keywords.length ? keywords : [""]) {
```
Zamiell, line 182: this is a bit arcane to read
is it more clear to just enclose the for loop in an if statement?

Zamiell: wait, is this code even doing anything? the loop won't be entered if keywords is empty

newduke: It is doing something, in fact exactly as the comment indicates. If the keyword list is empty, use the empty keyword to generate all possibilities. This code is needed for `noteHasMeaning` to work.

Zamiell: ok

PR #2133, opened 2021-03-14
Fix: make date&time column sortable closes #2130

### LINKED ISSUE #2130: Sorting games by Date & Time not working (when viewing all games with certain seed)
I noticed that sorting games (when viewing all games with certain seed) does not work as expected for Date & Time.

Steps to reproduce:
* go to https://hanab.live/seed/p3v16s14
* sort column `Date & Time` by clicking on it (I just assume that this should be possible as it works for all the other columns)
* table is not sorted by date & time, e.g. there are some games from 2019, 2020, 2021 but then again some from 2020 and 2019
* I would expect it to be sorted by timestamps

btw, really nice game/implementation ;)

nikolockenvitz:
To fix the issue described in #2130, I just needed to tell the tablesorter plugin how to parse the date and time values in this column. As explained in the comments, my current fix assumes that the displayed date and time is always in the format `YYYY-MM-DD — hh:mm:ss`. Based on my observation this is true for my local tests as well as for the official website. But I do not know whether this is implicitly (Go/Postgres/...?) or specified somewhere (I didn't found something corresponding). If the date and time string is in another format, sorting will not work (order as returned from backend; behavior is exactly the same as in current productive version).

What do you think? Is there a possibility that the date and time string will have another format? Is this fine or do you have any idea for a more generic approach?

### DIFF
--- a/server/src/views/history.tmpl
+++ b/server/src/views/history.tmpl
@@ -74,7 +74,7 @@
       <th># of Players</th>
       <th>Score</th>
       <th>Variant</th>
-      <th>Date & Time</th>
+      <th class="sorter-date-and-time">Date & Time</th>
       <th>Players</th>
       {{if not .SpecificSeed}}<th>Other Scores</th>{{end}}
       {{if eq .Title "Tagged Games" }}<th>Tags</th>{{end}}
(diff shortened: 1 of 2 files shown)


Zamiell:
the format comes from golang `time.Time`

this seems fine for now, thanks

Zamiell:
there's lots more to do if you want to squash more bugs =)

PR #2134, opened 2021-03-14
implemented /removetimer and /settimer pre-game commands

DrAnax:
pre-game chat commands:

/settimer [minutes] [seconds]
/removetimer

known "bug": doesn't check for speedruns


Zamiell:
i think what we want is a "Change Settings" button that brings up the normal UI, not custom commands that only expose partial-game-options-changing functionality.

the reason that i say this is that all of this code that you just submitted would be deleted in the event of having the former thing.

DrAnax:
ok, will go that way then

PR #2142, opened 2021-03-16
Added "clued" note

### LINKED ISSUE #2119: make x note for unclued

### LINKED ISSUE #2123: Add "clued" as a special note
From the features:
> A note of "unclued" can be written to manually remove the border that normally appears around a card when it is touched by one or more clues.

I'd love to see `clued` get added to the list of special notes. This would raise the card and give it spoOooOOooOOOky Halloween borders, like with any touched card. This will be really helpful for cards that are considered gotten and which affect finesse positions.

For example, after a shadow finesse, the target of the shadow is considered globally known and gotten – Bob should therefore ignore this card when counting finesse positions for an ejection. Conversely, if slot 1 is "gotten" by a finesse, an ejection would still call for slot 2. It'd be much easier for both clue-givers and receivers to track finesse positions in situations like these with a `clued` note.

newduke:
Fixes #2123
Fixes #2119


Zamiell:
nice, thanks

PR #2144, opened 2021-03-16
[client] set focus to close button, closes with [ENTER]

DrAnax:
When the user types an invalid command, a modal warning is shown. This fix moves the focus to the "close" button, which can now be closed by pressing ENTER.

After modal close, the focus is set to the last HTML element, which is the appropriate text box.

### DIFF
--- a/client/src/modals.ts
+++ b/client/src/modals.ts
@@ -88,7 +88,14 @@ export function warningShow(msg: string): void {
   globals.modalShowing = true;
 
   $("#warning-modal-description").html(msg);
-  $("#warning-modal").fadeIn(FADE_TIME);
+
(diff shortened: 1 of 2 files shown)


client/src/modals.ts
```diff
@@ -88,7 +88,14 @@ export function warningShow(msg: string): void {
   globals.modalShowing = true;
 
   $("#warning-modal-description").html(msg);
-  $("#warning-modal").fadeIn(FADE_TIME);
+
+  // store the screen's active element
```
Zamiell, line 92: comments must be capitalized

Zamiell:
nice, thank you

PR #2155, opened 2021-03-26
[client] allow < & > " in notes by using HTML entities

### LINKED ISSUE #2149: Notes: left angle bracket causes deletion of partial note
Steps to reproduce
1. Start writing a note
2. Write `hello <world`
3. Press enter
4. Look at the note
5. It contains `hello `

DrAnax:
Fixes #2149

### DIFF
--- a/client/src/game/ui/notes.ts
+++ b/client/src/game/ui/notes.ts
@@ -170,9 +170,9 @@ export function openEditTooltip(card: HanabiCard): void {
         );
       }
 
-      // Strip any HTML elements
+      // Convert &, <, >, " to HTML entities
       // (to be thorough, the server will also perform this validation)
-      newNote = stripHTMLTags(newNote);
+      newNote = convertHTMLEntities(newNote);
 
       set(card.state.order, newNote);
     }
@@ -219,7 +219,10 @@ export function setAllCardIndicators(): void {
   }
 }
 
-function stripHTMLTags(input: string) {
-  const doc = new DOMParser().parseFromString(input, "text/html");
-  return doc.body.textContent ?? "";
+function convertHTMLEntities(input: string) {
+  return String(input)
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;");
 }
--- a/client/src/game/ui/notes.ts
+++ b/client/src/game/ui/notes.ts
@@ -220,11 +220,7 @@ export function setAllCardIndicators(): void {
 }
 
 function convertHTMLEntities(input: string) {
-  return String(input)
-    .replace(/&/g, "&amp;")
-    .replace(/</g, "&lt;")
-    .replace(/>/g, "&gt;")
-    .replace(/"/g, "&quot;")
-    .replace(/\//g, "&sol;")
-    .replace(/'/g, "&apos;"); // cspell:disable-line
+  const p = document.createElement("p");
+  p.appendChild(document.createTextNode(input));
+  return p.innerHTML;
 }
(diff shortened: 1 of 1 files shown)


client/src/game/ui/notes.ts
```diff
@@ -170,9 +170,9 @@ export function openEditTooltip(card: HanabiCard): void {
         );
       }
 
-      // Strip any HTML elements
+      // Convert &, <, >, " to HTML entities
       // (to be thorough, the server will also perform this validation)
```
timotree3, line 174: This comment is now out of date right?
```suggestion
      // Convert &, <, >, " to HTML entities
      // (to be thorough, the server will also validate that it contains no HTML tags)
```

timotree3 (commented):
That was fast! Thanks for fixing this

Zamiell:
i'd prefer if you also did it on more characters
how about single quote, forward slash, both of those are used in xss attacks

client/src/game/ui/notes.ts
```diff
@@ -219,7 +219,10 @@ export function setAllCardIndicators(): void {
...
 
-function stripHTMLTags(input: string) {
-  const doc = new DOMParser().parseFromString(input, "text/html");
-  return doc.body.textContent ?? "";
+function convertHTMLEntities(input: string) {
+  return String(input)
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;");
```
timotree3, line 227: If you want to make this really robust it might make sense just to try putting the contents inside of a `p` and see how to browser escapes it. I tested that this works on Firefox.
```suggestion
function convertHTMLEntities(input: string) {
  const p = document.createElement("p");
  p.appendChild(document.createTextNode(input))
  return p.innerHTML
```

At that point I wonder why we don't just remove all this logic and do it where we're displaying the notes instead

Zamiell: is that safe

timotree3: why wouldn't it be? It doesn't interpret the text passed to `createTextNode` as HTML

DrAnax: It only converts &, <, >

Zamiell: https://stackoverflow.com/questions/6234773/can-i-escape-html-special-chars-in-javascript

anax is this where you found the code?

Zamiell: its probably safer to use jquery's test function.
its an antipattern to roll your own security

DrAnax: > https://stackoverflow.com/questions/6234773/can-i-escape-html-special-chars-in-javascript
> 
> anax is this where you found the code?

Nope, just wrote it

DrAnax: > its probably safer to use jquery's text function.
> its an antipattern to roll your own security

Can't use text() on an input element

Zamiell: can't you use it as a replacement for the function that you hand-crafted?

DrAnax: I can create a new element, paste the input's value and get it back via the text() function, is this ok?

Zamiell: oh, i guess that is pretty similar to what tim is proposing

DrAnax: So, use this?

function convertHTMLEntities(input: string) {
  const p = document.createElement("p");
  p.appendChild(document.createTextNode(input))
  return p.innerHTML

Zamiell: ok i looked it up and jquery's text() function just does the code that tim suggested, so let's go with tim's code instead

Zamiell: also, i am talking to you on discord, but you are not online =p

Zamiell:
nice work

PR #2160, opened 2021-03-28
Fix #2038

### LINKED ISSUE #2038: Split command args on newline
It is common for people to type
```
/replay 12345 12
Question text
```
which doesn't work currently.

NumerateClutter:
Only addresses /replay in discord, not in the lobby


NumerateClutter:
Oops, looks like the linter is failing -- let me address that and re-pull

Issue #2079, opened 2021-01-31
Aliases for commonly-used commands


Lanvin2219:
Hi, 

Would it be possible to add some aliases please?

Some ideas :

/m for /missing
/p for /playerinfo
/si for /startin
/sv for /setvariant
/fv for /findvariant
/f for /friend

PR #2164, opened 2021-04-06
 Aliases for commonly-used commands #2079 

plambros:


client/src/chatCommands.ts
```diff
@@ -212,6 +213,7 @@ function playerinfo(_room: string, args: string[]) {
     });
   }
 }
+chatCommands.set("p ", playerinfo);
```
Zamiell, line 216: why is there a space here

server/src/chat_command.go
```diff
@@ -72,6 +75,7 @@ func chatCommandInit() {
 	chatCommandMap["w"] = chatCommandWebsiteOnly
 	chatCommandMap["whisper"] = chatCommandWebsiteOnly
 	chatCommandMap["msg"] = chatCommandWebsiteOnly
+	chatCommandMap["f"] = chatCommandWebsiteOnly
```
Zamiell, line 78: why add it here if you are not going to add it to the website?

PR #2170, opened 2021-04-15
Add critical fours variants

### LINKED ISSUE #2168: Variant (Re)Proposal: Critical Fours
Fairly self explanatory, only one 4 of each suit.

This is nice as an alternative to dark suits for variants that impose higher efficiency requirements, since black suits come with a significant amount of luck around bottom-decks that is out of your control. Crit 4s gives a similar efficiency bar but the bottom-deck with normal rules still lets you get N-1 points and with bottom deck blind plays it is very unlikely to prevent a perfect on its own.

I have implemented this on a fork and we've played a game of it. Nothing too special IMO but the players mostly agreed that they prefer it to dark suits. https://youtu.be/Rs22PPtL3lI

bhamrick:
Closes #2168


bhamrick:
Resolved the conflicts and rebased the branch.

Zamiell:
thanks

Zamiell:
i pushed this to prod and apparently it wasn't giving any 4s in normal variants, did you test this before submitting the PR?

bhamrick:
I did test a critical fours game, did not think to test a normal game. I'll take a look to see what went wrong.

bhamrick:
Ok, I just made an oversight that the default `amountToAdd` was set by the final else clause rather than beforehand, and adding the new branch made that else clause not execute. I pushed a fixed version to my branch but I don't see a way to update this pull request so I guess I make a new one?

Zamiell:
yep

PR #2174, opened 2021-04-22
feat: allow wildcard characters in the /seed endpoint

aliblong:
By simply using the SQL LIKE comparator in the backend, the user can request results across multiple seeds, through use of wildcard metacharacters.

The use case driving this feature addition is the NoVarathon initiative, where we aim to collect results for seeded no variant games in close to real time. The seed name will follow a pattern, and we want results for all seeds matching that pattern.

### DIFF
--- a/server/src/models_games.go
+++ b/server/src/models_games.go
@@ -321,8 +321,12 @@ func (*Games) GetGameIDsUser(userID int, offset int, amount int) ([]int, error) 
 	return gameIDs, nil
 }
 
-func (*Games) GetGameIDsSeed(seed string) ([]int, error) {
+const PageSize = 1000
+
+func (*Games) GetGameIDsSeed(seed string, page uint) ([]int, error) {
 	gameIDs := make([]int, 0)
+    record_number_lower_bound := page * PageSize
+    record_number_upper_bound := (page + 1) * PageSize
 
 	// Neither % nor _ chars are permitted in seed names, so the change in comparator
 	// from '=' to 'LIKE' should not affect any previously working queries.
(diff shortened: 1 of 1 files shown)


Zamiell:
"Someone can in principle select all games ever played by simply specifying '%' as the seed pattern"

that sounds like it would ddos production

aliblong:
Is your main concern about malicious or accidental DOS? It seems like it wouldn't have much potential for abuse, and like it would be relatively straightforward to address that type of abuse if it becomes an actual problem. But like I say in the comment, I can add a limit clause to the query if you think that's prudent.

Zamiell:
> Is your main concern about malicious or accidental DOS?

yes, right now the server chokes on getting p4v0s0, for example

aliblong:
When you say "chokes", do you mean that it's configured such that there's an upper limit on the response size? When I query for that seed, I get back an empty response; when I query for `p4v0s5`, I get an instantaneous response with 1456 games.

Zamiell:
there *should* be an upper limit on the response size, but there isn't, which is what causes the issue,
a PR to fix that would be welcome - specifically, the end result being an error message displayed to the user and the server not DOSing itself. not really sure what the best way to do it is though.
if that was in place, then we could just accept this current PR as-is.

Zamiell:
in principle, the server should probably not allow requests for more than 1000 games/seeds/whatever at a time

Zamiell:
in a hypothetical where the server has a 1000 game cap, then one would get the 1456 results for `p4v0s5` with two separate queries:
- `/seed/p4v0s5?1-999`
- `/seed/p4v0s5?1000-2000`

PR #2177, opened 2021-04-25
Show suits tooltips

### LINKED ISSUE #2073: In-game explanation of the used variant
If I'm a new player, and started e.g. spectating a game of some variant, then I'll have no idea what is this "light pink" or "muddy rainbow" suit is. So I'd move the mouse to that suit, and there would be a tooltip that explains that "light pink is ...."

And such tooltip should mention which letter I should use in the notes to mark my card as of that suit, like "3y" (yellow 3) but for some other suit. Such letter is currently only visible in the colorblind mode.

DarthGandalf:
Fix #2073

### DIFF
--- a/client/src/game/types/CardState.ts
+++ b/client/src/game/types/CardState.ts
@@ -1,6 +1,5 @@
 import Color from "./Color";
 
-// "playStack" is an empty play stack, "deck" is a play stack with some cards on it
 type CardLocation = "deck" | "discard" | "playStack" | number;
 
 export default interface CardState {
--- a/client/src/game/types/CardState.ts
+++ b/client/src/game/types/CardState.ts
@@ -1,5 +1,6 @@
 import Color from "./Color";
 
+// "playStack" is an empty play stack, "deck" is a play stack with some cards on it
 type CardLocation = "deck" | "discard" | "playStack" | number;
 
 export default interface CardState {
--- a/client/src/game/reducers/initialStates/initialCardState.ts
+++ b/client/src/game/reducers/initialStates/initialCardState.ts
@@ -1,3 +1,4 @@
+import * as deck from "../../rules/deck";
 import CardState from "../../types/CardState";
 import Variant from "../../types/Variant";
 
@@ -15,9 +16,11 @@ export default function initialCardState(
     possibleRanks.forEach((r) => possibleCards.push([s, r])),
   );
 
+  const total = deck.totalCards(variant);
+
   return {
     order,
-    location: "deck",
+    location: order < total ? "deck" : "playStack",
     suitIndex: null,
     rank: null,
     possibleCardsFromClues: possibleCards,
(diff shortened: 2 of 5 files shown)


client/src/game/types/CardState.ts
```diff
@@ -1,5 +1,6 @@
 import Color from "./Color";
 
+// "playStack" is an empty play stack, "deck" is a play stack with some cards on it
```
Zamiell, line 3: its not necessarily empty

Zamiell:
fails CI, make it pass CI

Zamiell:
does this only affect spectators of ongoing games?

client/src/game/types/CardState.ts
DarthGandalf: Then why there is a need for two?

DarthGandalf:
No, it works for players too

Zamiell:
that sounds very distracting for players, to have tooltips pop up constantly as they hover over cards when considering to play them

client/src/game/types/CardState.ts
Zamiell: what do you mean two?

DarthGandalf: What do these values mean? In my tests I needed to handle both: one for case of empty stack, another for case of something on it.
What's the intended difference?

DarthGandalf:
> that sounds very distracting for players, to have tooltips pop up constantly as they hover over cards when considering to play them

Do you prefer these tolltips somewhere else instead of on stacks?

DarthGandalf:
> as they hover over cards when considering to play them

How do you play the card which is played already? The tooltip is not shown for cards in hands

Zamiell:
```
  isOnPlayStack(): boolean {
    const { location } = this.state;
    return location === "playStack" || location === "deck";
  }
  ```
i dont understand this function, if the location is in the deck then why would it be on a pack stack?

Zamiell:
> Do you prefer these tolltips somewhere else instead of on stacks?

do they appear on the cards or just on an empty stack?

DarthGandalf:
> i dont understand this function, if the location is in the deck then why would it be on a play stack?

I'm wondering the same. I don't understand what exactly "deck" means here. From what `state.location` was returning, we need these 2 values here. One is for the empty play stack, another for a play stack with some cards on it.

DarthGandalf:
This is why I had to add a comment to the enum.

Zamiell:
i think you uncovered a real bug in the website then
we will need to investigate why it says "deck" if the real  location is the play stack
that function "isOnPlayStack()" should be deleted and we should fix this other bug first

Zamiell:
can you look into this?

DarthGandalf:
It probably is set in src/game/reducers/initialStates/initialCardState.ts and never updated?

I don't know what this code should do and how the website works in general.

DarthGandalf:
I removed the function. Now the tooltip is shown only on non-empty stacks.

I'll leave it to you to fix the bug how you see fit.

Zamiell:
looks good now thanks

PR #2183, opened 2021-05-08
Add ambiguous/dual up/down variants

DarthGandalf:


DarthGandalf:
Selection_045

Zamiell:
up and down is only intended to be combined with basic suit variants

DarthGandalf:
> up and down is only intended to be combined with basic suit variants

why?

Zamiell:
because if you allow every permutation then the number of variants will get out of hand

PR #2184, opened 2021-05-09
Make name of the variant more visible

DarthGandalf:
I often see that players notice that 1s or 5s are e.g. rainbow only by
the middle of the game

### DIFF
--- a/client/src/game/ui/Elements.ts
+++ b/client/src/game/ui/Elements.ts
@@ -49,6 +49,8 @@ export default class Elements {
   deckTurnsRemainingLabel1: Konva.Text | null = null;
   deckTurnsRemainingLabel2: Konva.Text | null = null;
   deckPlayAvailableLabel: Konva.Rect | null = null;
+  variantLabel: FitText | null = null;
+  variantUnderline: Konva.Line | null = null;
 
   // Extra elements on the right-hand side + the bottom
   clueLog: ClueLog | null = null;
--- a/client/src/game/ui/drawUI.ts
+++ b/client/src/game/ui/drawUI.ts
@@ -437,10 +437,10 @@ function drawDiscardStacks() {
   // Local variables
   let discardStackSpacing;
   if (globals.variant.suits.length === 6) {
-    discardStackSpacing = 0.04;
+    discardStackSpacing = 0.038;
   } else {
     // 3, 4, or 5 stacks
-    discardStackSpacing = 0.05;
+    discardStackSpacing = 0.047;
   }
 
   for (let i = 0; i < globals.variant.suits.length; i++) {
@@ -449,9 +449,9 @@ function drawDiscardStacks() {
     // Make the discard stack for this suit
     const discardStack = new CardLayout({
       x: 0.81 * winW,
-      y: (0.61 + discardStackSpacing * i) * winH,
+      y: (0.64 + discardStackSpacing * i) * winH,
       width: 0.17 * winW,
-      height: 0.17 * winH,
+      height: 0.15 * winH,
       listening: false,
     });
     globals.elements.discardStacks.set(suit, discardStack);
@@ -1321,7 +1321,7 @@ function drawStatistics() {
     x: clueLogValues.x * winW,
     y: 0.53 * winH,
     width: clueLogValues.w! * winW,
-    height: 0.06 * winH,
+    height: 0.09 * winH,
     fill: "black",
     opacity: 0.2,
     cornerRadius: 0.01 * winW,
@@ -1329,10 +1329,49 @@ function drawStatistics() {
   });
   globals.layers.UI.add(statsRect);
 
+  const variantLabel = new FitText({
+    text: globals.variant.name.replace(/\s*\(\d Suits\)/, ""),
+    x: 0.825 * winW,
+    y: 0.54 * winH,
+    width: 0.15 * winW,
+    height: 0.03 * winH,
+    fontSize: 0.02 * winH,
+    fontFamily: "Verdana",
+    fill: LABEL_COLOR,
+    shadowColor: "black",
+    shadowBlur: 10,
+    shadowOffset: {
+      x: 0,
+      y: 0,
+    },
+    shadowOpacity: 0.9,
+    align: "center",
+    listening: false,
+  });
+  globals.layers.UI.add(variantLabel);
+  globals.elements.variantLabel = variantLabel;
+
+  const variantUnderline = new Konva.Line({
+    points: [0.825 * winW, 0.56 * winH, 0.965 * winW, 0.56 * winH],
+    stroke: LABEL_COLOR,
+    strokeWidth: 1,
+    lineJoin: "round",
+    shadowColor: "black",
+    shadowBlur: 5,
+    shadowOffset: {
+      x: 0,
+      y: 3,
+    },
+    shadowOpacity: 0,
+    listening: false,
+  });
+  globals.layers.UI.add(variantUnderline);
+  globals.elements.variantUnderline = variantUnderline;
+
   const paceTextLabel = basicTextLabel.clone({
     text: "Pace",
     x: 0.825 * winW,
-    y: 0.54 * winH,
+    y: 0.57 * winH,
     fontSize: 0.02 * winH,
     listening: true,
   }) as TextWithTooltip;
@@ -1352,7 +1391,7 @@ function drawStatistics() {
   const paceNumberLabel = basicNumberLabel.clone({
     text: "-",
     x: 0.9 * winW,
-    y: 0.54 * winH,
+    y: 0.57 * winH,
     fontSize: 0.02 * winH,
     listening: true,
   }) as Konva.Text;
@@ -1372,7 +1411,7 @@ function drawStatistics() {
   const efficiencyTextLabel = basicTextLabel.clone({
     text: "Efficiency",
     x: 0.825 * winW,
-    y: 0.56 * winH,
+    y: 0.59 * winH,
     fontSize: 0.02 * winH,
     listening: true,
   }) as TextWithTooltip;
@@ -1402,7 +1441,7 @@ function drawStatistics() {
   const efficiencyNumberLabel = basicNumberLabel.clone({
     text: "-",
     x: 0.9 * winW,
-    y: 0.56 * winH,
+    y: 0.59 * winH,
     fontSize: 0.02 * winH,
     listening: true,
   }) as TextWithTooltip;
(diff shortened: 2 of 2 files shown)


DarthGandalf:
On the deck, which has the tooltip of the full variant name

DarthGandalf:
Selection_058

DarthGandalf:
> maybe the ones should be drawn differently

That's an interesting option, yeah. But that doesn't solve case when there are no visible 1s (or 5s) currently, while that short string is there during the whole game

Zamiell:
> But that doesn't solve case when there are no visible 1s (or 5s) currently, while that short string is there during the whole game

i could make the same argument for the variant called "Omni (6 Suit)"

"It doesn't solve the case where there are no visible Omni cards, while that short string is there during the whole game"

DarthGandalf:
> It doesn't solve the case where there are no visible Omni cards

But there's always a visible stack for the Omni cards, even if it's just the stack base. So it's hard to not notice it.

Zamiell:
ok, then i could make the same argument for the variant called "Throw It in a Hole (6 Suit)", which has no other visible indicators that the rules of the game are changed

DarthGandalf:
> i could imagine an alternating clues game

I wouldn't be opposed to adding such indication

Zamiell:
if we want to draw the name of the variant for all variants, there isn't a real good place to add it
(which is why we made it on a tooltip)

DarthGandalf:
> (which is why we made it on a tooltip)

That tooltip hardly anyone looks at, and I saw such confusion in games. If you don't like such abbreviation of the parts of the variant which are not visible elsewhere (what I did in this PR), I can try moving the whole variant name to some other place if you prefer.

Zamiell:
i'm open to suggestions for where you would want to display the full variant name, the problem is I don't think there is a very good option

as for displaying the variant abbreviation, you would have to write a custom system to make abbreviations for every variant, and that might be a pain to maintain

Zamiell:
> What about a top right corner above the list of clues?

show a ss

> Anything is missing from it other than alternating clues?

from what you wrote already? it looks like you only handled special ranks and nothing else

DarthGandalf:
>  it looks like you only handled special ranks and nothing else

Because most other things are already visible somewhere...

vEnhance:
> if we want to draw the name of the variant for all variants, there isn't a real good place to add it
> (which is why we made it on a tooltip)

what if we make the log a couple of lines shorter and put it there at upper left

i don't know about others, but i never use more than the last few lines of the log, if i need to look back further i'll usually just rewind anyways

mockup

Zamiell:
> what if we make the log a couple of lines shorter and put it there at upper left

remember that there are 2 modes.
offer suggestions for keldon mode first since that has less space, BGA mode isn't as relevant

vEnhance:
ah right i keep forgetting keldon mode exists (i can't get used to it :smile: )

yeah i take back my suggestion, upper right is probably best then if we want it for both modes

(would it be too silly to have a feature for bga mode and not keldon mode)

Zamiell:
>  I'm thinking White-Ones could have a ❌1 next to the color clues and Rainbow-Ones could have a ✔️1 or something like that.

sounds kind of cluttery to be honest.
and there isn't a lot of room there.
show a proposed screenshot in keldon mode with all of the hypothetical buttons showing

DarthGandalf:
> another thing to note is that the clue count already has to truncate when there are too many clues, so we need to be very careful about using up space here, or else it causes more truncation

This is why I tried to put the abbreviation on the deck instead, which has some space. I chose deck because it's the deck's tooltip which has the full name.

> you're always going to see the clue UI before you take your first turn

This is true, however, for example, I don't look at the selection of buttons, and based on that choose which clue to give to whom. Instead, I first decide to give this kind of finesse to get those cards played, which would be... ok, G to James, and then select the button G and the player James. I wouldn't necessarily notice that there's no button for "1" unless I specifically try to give that clue (and fail)

Zamiell:
> This is why I tried to put the abbreviation on the deck instead

right, so given the assumption that we want the name of the variant to always be displayed, consistently for every variant, then using the deck has 2 problems:

1) raw text on the deck is pretty fuckin ugly
2) there's probably not enough room to put the full variant name on the deck, which means we need to write (and maintain) a variant abbreviation system that generates abbreviations for every possible variant and every future variant

DarthGandalf:
> given the assumption that we want the name of the variant to always be displayed

This I disagree with.

The full name is already available, in the tooltip. And most parts of the variants (additional suits etc) are already visible in some way on the board.

> system that generates abbreviations for every possible variant

Therefore I'm repeating my question: which variants are still missing? Alternating clues, anything else?
Since the abbreviation system needs only to show things which are not immediately visible elsewhere, the "if" blocks I added already do most of the job.

DarthGandalf:
> raw text on the deck is pretty fuckin ugly

Yeah, it's not ideal.

But showing the variant name directly without any transformation system in place which would show it differently would still require showing raw text. Somewhere.

Zamiell:
> The full name is already available, in the tooltip. And most parts of the variants (additional suits etc) are already visible in some way on the board.

I don't want to create a variant-reminding feature that is hard coded to only display when playing Special-Ones, or Special-Fives, or Alternating Clues, or Color Blind, or Number Blind, or Totally Blind, or Color Mute, or Number Mute, or Clue Starved, or Cow & Pig, or Duck, or Reversed, or Throw it in a Hole, or Up and Down.

such a system would be

1) inconsistent and jarring because it shows sometimes and not otherwise
2) subjective in that some variants are arbitrarily chosen as to be worth reminding the player about
3) prone to failure because when new variants are added, you would have to also add them to the new custom system

thus, I think that in order to correctly alleviate the problem of players forgetting what variant they are playing, you should make a UI element that shows the variant consistently across all situations.
and if we can't come up with a prototype UI element that looks good and feels right, then we should leave the variant on the tooltip as-is

timotree3:
What about an automatic chat message after the game starts "Reminder: White-Ones"

DarthGandalf:
Selection_062
Selection_061

WDYT?
I had to slightly move the hands in 3p/5p keldon modes, but seems to work

Zamiell:
it can't go there because thats where the hand is in 5p

DarthGandalf:
No, the hand is above it

Selection_063

Zamiell:
that looks cramped and pretty bad

Zamiell:
also it looks like it has something to do with test2

Zamiell:
it looks cramped and bad in BGA mode as well
show a ss with the clue ui

DarthGandalf:
Selection_064

Zamiell:
the floating text looks pretty bad, also it is crammed up against the replay button, what about if you made it a real ui element like how pace & eff look like, with the same spacing that they have

DarthGandalf:
Okay, here's another design, which should work for keldon mode too (near Pace/Efficiency): 

Selection_066
Selection_065

Zamiell:
not in love with that, maybe it would look better if it was separated into its own element and placed above pace/eff

Zamiell:
that's true, you could try putting it above pace in the same element and draw a small horizontal line

Zamiell:
looks pretty bad imo, others can chime in

DarthGandalf:
Selection_068

DarthGandalf:
Selection_069

padiwik:
I like these latest screenshots, though the variant/pace/eff box should be made taller (eating into the clue log). This is also a great place because it's one of the places a strong player checks at the beginning of the game to see if it's a hard variant and what pace is needed.

Aigul9:
how will “muddy-rainbow-ones and muddy rainbow” look like?

DarthGandalf:
Selection_072

Zamiell:
can we programmatically determine the longest variant name and then test that?

DarthGandalf:
What do you mean? This is one-time testing, why do you want to determine it programmatically?

And if you want an automatic testing, what exactly you want to test?

Zamiell:
im just talking about showing a ss with the longest variant name, because the variant that valetta suggested isn't even a real variant

DarthGandalf:
> because the variant that valetta suggested isn't even a real variant

How so? Where does my screenshot come from then? You think I forged it?

Check existance of that variant yourself on the live website.

Zamiell:
the variant of "Muddy-Rainbow-Ones & Muddy Rainbow (6 Suits)" is a real variant, but not the one that you posted

DarthGandalf:
That's because I'm removing the (N Suits) suffix. It only consumes space without adding any value, since you can count number of suits

Zamiell:
does a variant exist with a longer name exist than "Muddy-Rainbow-Ones & Muddy Rainbow"?

Aigul9:
muddy-rainbow-fives))

DarthGandalf:
Selection_073

Zamiell:
is that the longest possible variant that exists?

DarthGandalf:
please show a longer one if you disagree. The `FitText` class should handle it anyway

Zamiell:
im asking, can we programmatically determine the longest variant name and then test that?

Zamiell:
specifically without the suit suffix, since you are removing that

DarthGandalf:
```
$ python
Python 3.8.10 (default, May  5 2021, 10:34:28) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> import json
>>> max((re.sub(r'\s*\(\d Suits\)', '', variant['name']) for variant in json.load(open('variants.json'))), key=len)
'Muddy-Rainbow-Fives & Muddy Rainbow'
>>> 
```

Zamiell:
the latest screenshot looks okay to me, what do other people think

vEnhance:
I like the latest screenshot as well!

padiwik:
Could you share a discard pile from 6 suits? Like this game https://hanab.live/export/528506 . I hope it's not too cramped with this change.

Also, I think the vertical space between the horizontal line and "Pace" should be less.

DarthGandalf:
Selection_075

DarthGandalf:
> Also, I think the vertical space between the horizontal line and "Pace" should be less.

Selection_076

padiwik:
I love it!

DarthGandalf:
> i think this one looks better

Feel free to merge only the first commit from this branch. The second one decreased the distance.

Zamiell:
i dont know how to do that, can u just revert the 2nd commit?

DarthGandalf:
Done

Zamiell:
bikeshedding is complete, i will merge this now, thanks

in the future, when u want to change the website, it might be better to open issues first before you go through the work of coding up a PR
that way everyone can discuss them, and you avoid wasting your own time if the issue/PR will be rejected

padiwik:
The reason I prefer less spacing is because the asymmetry above and below the line looks weird with short variant names (https://user-images.githubusercontent.com/325092/118092968-ad74b180-b3c4-11eb-9a58-d2e6acdc28ff.png). There's probably some way to make the line always be equidistant based on the height of the `FitText`.

PR #2186, opened 2021-05-10
DDA indicators using cardRules.canPossiblybe 

vEnhance:
This is a client-side implementation of a DDA indicator to make life a bit easier for dummies like me that keep forgetting DDA exists :stuck_out_tongue: 

When *you* are in DDA, it will put a cyan question mark next to any card that could be the card that was just discarded:
in-dda

When *someone else* is in DDA, it will put a cyan question mark for every non critical card that could be the card that was just discarded (so, if a player is chop moved, that won't affect our calculation):

Screenshot_2021-05-10_16-07-30

The indicator use `cardRules.canPossiblyBe` so as to be maximally intelligent. That means for example

* If Alice discards r4, and Bob has negative 4 information on his chop, then the cyan indicator will *not* appear for any player, indicating to the entire team that Bob is not actually in DDA.
* If Alice discards r4, and Bob sees the other copy of r4 in Cathy's hand, then the cyan indicator will be appear for both Alice and Bob, but *not* for Cathy (who is still unaware she has r4). Hence, if Bob goes on to discard anyways, Cathy will be surprised and mark her hand for r4 elimination notes.

Screenshot_2021-05-10_16-13-17

Indicators never appear on cards that are clued. These indicators only appear if Hyphenated options are enabled (to avoid confusing new players with a hard-to-explain convention).


Zamiell:
what are the yellow things

timotree3:
They're hats. They're not part of this PR; they're from a hat guessing ruleset enforcement extension.

client/src/game/types/StatsState.ts
```diff
@@ -24,7 +24,8 @@ export default interface StatsState {
   // (future efficiency is simply "cardsNotGotten / cluesStillUsable")
 
   // Other
-  readonly doubleDiscard: boolean;
+  // store the order of the double-discard candidate, or null if not in DDA
```
Zamiell, line 27: comments should be capitalized

timotree3:
I think you should only include things in the state if you're going to use it to generate future state. If it's just a function of the current state then it's fine for it to go in a view like in this PR.

timotree3:
Actually it looks like it's not in a view.

Zamiell:
if it isn't in state then isn't it susceptible to bugs?
e.g. look at the bug that anax just fixed with the shadows.
the fix was to add shadows to state

timotree3:
The shadows were using a boolean state in the card object and were mutating that state on click and drag and stuff. This feature has a boolean state in the card object too, but it's reset on every action due to `setStatus` and it's completely determined by the redux state and it's never mutated directly, so it's not a problem I don't think.

Zamiell:
what are the performance implications of having it in state?
having some variables exist in state and some variables exist on UI objects seems confusing.

timotree3:
There's not even a variable on the UI object, it's just whether the indicator is visible or not. If we added it to the state, the code on HanabiCard will be the same, just instead of setting the visibility of the indicator based on 

```
    this.setDDA(
      dda !== null &&
        this.state.location === visibleState.turn.currentPlayerIndex &&
        status !== CardStatus.Critical &&
        cardRules.canPossiblyBe(
          this.state,
          visibleState.deck[dda].suitIndex,
          visibleState.deck[dda].rank,
        )
    );
```

we would move that logic to a reducer and HanabiCard would have

```
this.setDDA(this.state.shouldShowDDAIndicator);
```

you still have the same boolean state in the cards (visible or not) and it's still determined by the Redux state. You've just added book-keeping and redundancy.

Zamiell:
1) evan put a message here when you push new commits so that it goes to discord and i am notified
2) it looks like you added it to the cardstate.
but notice that CardStatus is not on cardstate and is instead an array of gamestate.
tim/whoever, do we know why CardStatus is that way instead of a subproperty of the cardstate?
and depending on the answer to that question, it might be better to put evan's boolean on gamestate as well

vEnhance:
>  tim/whoever, do we know why CardStatus is that way instead of a subproperty of the cardstate?

Seems like `state.cardStatus` indexed by `(suit, rank)` whereas `state.deck` is indexed by `order`. So e.g. in no variant, there are 25 card statuses and 50 card states.

DDA has to be based on order since it depends on where the card is / empathy / etc. Card status doesn't have to be so it makes sense it would be tracked separately rather than duplicating the information.

(Correct me if I am wrong here, this was just the impression I got.)

Zamiell:
in your PR, can you some comments to cardstatus on gamestate.ts that describe why this isn't on cardstate
and also maybe some comments on your new entry in cardstate.ts that explain why it is not congruent with cardstatus
other than that it looks good

PR #2189, opened 2021-05-11
Make hard variant and low score phase indicators colorblind friendly

nitawa-hl:
When a variant is hard or it's the low score phase, the numbers of the corresponding info are colored.
This PR makes the font also bold if colorblind mode is activated.

colorblind


Zamiell:
looks good thanks

PR #2195, opened 2021-05-16
Show the current efficiency

DarthGandalf:
Selection_077

### DIFF
--- a/client/src/game/types/ReplayArrowOrder.ts
+++ b/client/src/game/types/ReplayArrowOrder.ts
@@ -12,7 +12,8 @@ enum ReplayArrowOrder {
   Strike2 = -8,
   Strike3 = -9,
   Pace = -10,
-  Efficiency = -11,
+  CurrentEfficiency = -11,
   MinEfficiency = -12,
+  FutureEfficiency = -13,
 }
 export default ReplayArrowOrder;
--- a/client/src/game/ui/Elements.ts
+++ b/client/src/game/ui/Elements.ts
@@ -55,9 +55,11 @@ export default class Elements {
   // Extra elements on the right-hand side + the bottom
   clueLog: ClueLog | null = null;
   paceNumberLabel: Konva.Text | null = null;
-  efficiencyNumberLabel: TextWithTooltip | null = null;
-  efficiencyPipeLabel: Konva.Text | null = null;
-  efficiencyMinNeededLabel: Konva.Text | null = null;
+  efficiencyCurrentNumberLabel: TextWithTooltip | null = null;
+  efficiencyPipe1Label: Konva.Text | null = null;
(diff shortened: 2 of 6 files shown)


Zamiell:
we already decided against this.
there's no reason to need to see the current efficiency

DarthGandalf:
The reason is the confusing label: if we are not efficient, the text says "efficiency: 4.00", which means we are very efficient, right?And "efficiency: 0.3" means not efficient at all

Zamiell:
Efficiency is short for Future Efficiency, so yes it is a confusing label
i'm open to suggestions if you want to rename the label, but cluttering the screen with even more useless numbers probably doesn't make the situation any better

DarthGandalf:
Maybe "Future Efficiency"?

Zamiell:
is there room for that?
it would push the numbers far off to the right side

DarthGandalf:
"Required efficiency" fits if to decrease the margin on sides

DarthGandalf:
Selection_078

Zamiell:
i dont think that looks particularly good

DarthGandalf:
Me neither, but still better than the confusing label

timotree3:
I think a much more important number for telling you how efficient to be is `remaining clues - cards not gotten`. If it less than the player count then you might not have enough clues for stalling, if it exceeds the player count then you can waste clues freely. You could call that number "spare clues" or something like that. You could also display it negated (`cards not gotten - remaining clues`) and label it "req. finesses" or something like that.

Zamiell:
i agree with that

nitpick: req finesses isn't a good label because it is very common to e.g. give a 2-for-1 normal clue

"Spare Clues" is also significantly shorter than "Required Efficiency", so that solves the UI problem as well

Zamiell:
what specific amount of spare clues is equivalent to 1.25 req eff?

timotree3:
3p and 4p single dark 5 suits (the canonical hard variant) both have starting spare clues of -5

DarthGandalf:
This number will probably be different for 3 suits and 6 suits

Zamiell:
the reason i ask is because i update to update the documentation everywhere, which is a pain

timotree3:
The hard variant definition could still use req eff if we displayed spare clues and req eff side by side. Or we could define hard variants as "check if the spare clues count is orange" and maintain the logic only in the website.

DarthGandalf:
Then it will fill arbitrary whether the variant is hard or not

DarthGandalf:
> if we displayed spare clues and req eff side by side

That's still having more numbers to show than now?

Zamiell:
> The hard variant definition could still use req eff if we displayed spare clues and req eff side by side.

i don't want to do that, see previous discussion in this thread

Issue #2197, opened 2021-06-06
Allow precluing color on rainbow cards during speedruns


RaKXeR:
Currently, if you attempt to left-click a rainbow card to preclue color, the client will send the color value "-1" to the server, which obviously fails.
It used to default to the leftmost color, but I believe the most helpful solution is to replace the -1 value with whatever color is currently selected in the interface, which can be selected using QWERTY.

I'm not at all up-to-date on the codebase, specially the client-side typescript, but a botchy implementation could be putting a verification and replacement of the value somewhere [here](https://github.com/Zamiell/hanabi-live/blob/22cdbe0379bfb27d606b4ec9c1dda2f4c6f2abdc/client/src/game/ui/turn.ts#L46). I couldn't actually find where the premove itself was generated.

Zamiell:
submit a pr

PR #2199, opened 2021-06-12
Fix left click color clues in speedruns

RaKXeR:
* Clue correct color when left clicking rainbow cards, properly taking into account the state of the clue interface
* Intentionally send invalid clue when color cluing truly colorless card
#2197


Zamiell:
lgtm

Issue #2078, opened 2021-01-31
Chat doesn't auto-scroll


mdavidson29:
![image](https://user-images.githubusercontent.com/30240608/106394104-b4e0f380-63af-11eb-8150-d20d2594a693.png)

The chat doesn't auto-scroll if it is already open. 

- Closing the chat and reopening it brings it to the bottom
- If new messages are sent while the chat is open, it does not auto-scroll down to show new messages

Zamiell:
the chat won't auto scroll if it isn't already set to be at the bottom-most line.
try scrolling all the way up and then scroll all the way to the bottom

mdavidson29:
Even if I scroll all the way to the bottom, it only auto-scrolls for 1 message.  After that, it doesn't auto-scroll.

mdavidson29:
We are doing a text review and I have to be constantly scrolling down.

Zamiell:
try on version 4966

PR #2208, opened 2021-07-04
Fix verification if chat needs to scroll

hello-hanabi:
I encountered this issue with the auto-scroll:

https://user-images.githubusercontent.com/86917256/124386937-a25b4500-dcd4-11eb-905d-dd5fd4990aaa.mov

It happens only when my browser window is not in full-screen and it fixes itself if I refresh the page. I'm on a MacOS 11.3.1
This could be the issue described in #2078

Basically I noticed that the `scrollHeight - scrollTop` value is 1px bigger than `clientHeight` and it causes the issue. I'm not sure where this pixel comes from (I initially thought it might be the border since it's 1px wide but it's not) so I decided to give it a +-2 px margin to decide if it should auto-scroll. This way it will also work when `scrollTop` is fractional sometimes.

Note: I didn't manage to test this change locally

### DIFF
--- a/client/src/chat.ts
+++ b/client/src/chat.ts
@@ -595,13 +595,10 @@ export function add(data: ChatMessage, fast: boolean): void {
   // Find out if we should automatically scroll down after adding the new line of chat
   // https://stackoverflow.com/questions/6271237/detecting-when-user-scrolls-to-bottom-of-div-with-jquery
   // If we are already scrolled to the bottom, then it is ok to automatically scroll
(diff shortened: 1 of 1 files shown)


Zamiell:
1) test the change locally
2) make it pass ci

hello-hanabi:
I tested it locally and the issue doesn't occur anymore! Also fixed the build (sorry I missed the spell checker)

Zamiell:
thanks, what is your id?

PR #2212, opened 2021-07-07
Fix score being permanently bold in colorblind mode

nitawa-hl:
Fixes (me-made) bug were score is always bold in colorblind mode, instead of being bold only during low score phase.


Zamiell:
thanks!

PR #2213, opened 2021-07-08
Show start rank pip

### LINKED ISSUE #2202: Up or Down: Show on the backs of cards whether a card could be a Start card
In Up or Down, show on the backs of cards whether the card could be a Start card, for example, by showing an S to the left of the 1:
image

Even though users cannot directly clue Start cards, they would still be `X`'d out if the user can deduce that the card is not a Start card because they can see all the Start cards matching the possible suits

With this change, in the screenshot above the rightmost card is clearly the Green Five. Without this, the user has to remember to check whether it could also be the Green Start.

DarthGandalf:
Fix #2202
Selection_088
Selection_089

### DIFF
--- a/client/src/game/ui/HanabiCard.ts
+++ b/client/src/game/ui/HanabiCard.ts
@@ -650,10 +650,6 @@ export default class HanabiCard
     }
 
     for (const [rank, pipState] of rankPipStates.entries()) {
-      if (rank > 5) {
-        // Don't show the pip for START cards (in "Up or Down" games)
-        continue;
-      }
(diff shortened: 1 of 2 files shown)


Zamiell:
nice work

timotree3:
Would it be possible to make it taller?

Zamiell:
it would probably look better as a capital S, yeah

DarthGandalf:
> Would it be possible to make it taller?

I'd like it to be taller. But for some strange reason (probably related to the font) when I use S instead of s, it doesn't render anything with font size 63. But it renders S fine with a smaller font size.

Issue #2222, opened 2021-07-16
DDA detection still plays the double-discard sound even when it shouldn't


mdavidson29:
The new code that shows question marks when cards are at risk of double-discards isn't properly integrated with the double-discard sound.  

As an example, the 4-tone double-discard sound was played even when g4 was discarded immediately after r4, despite having negative-red information on this card.  
![2](https://user-images.githubusercontent.com/30240608/125898809-505f1615-212d-4a21-b813-3bb9cbea7f0f.jpg)
![1](https://user-images.githubusercontent.com/30240608/125898812-06cb5c76-1d6a-468e-9a2b-349d9e40f512.jpg)

vEnhance:
I think there's a question of when exactly we want the sound to play, because the question marks depend on asymmetric information.

For example: Alice discards g4. Bob sees the other g4 in Cathy's hand, so Bob discards anyways, and Cathy marks her hand for g4 elimination notes.

The four-tone sound certainly plays for Cathy. The question is whether it should also play for Alice and Bob, to remind the team that elimination notes should be written.

mdavidson29:
That's a really good question!  I think it makes sense for it to play for Alice and Bob as well, because that discard meant something extra.  

So I think here are the cases:
- If it's universally known to be a safe discard (e.g. the game I showed above), then the tone should not play
- If it's not universally known to be a safe discard (e.g. the situation you just described), the tone should play for everyone

Does that make sense?

Zamiell:
both the sounds and the question mark icons are intended to be based on globally known information only

vEnhance:
Man, my memory is real awful, but I finally remembered the reason why I kept the status quo for sounds. The answer is: **chop moves**.

In the screenshot you showed, the client doesn't know if the green card was chop moved or not, which would affect DDA determination. So, that's why when I designed the question marks, I decided explicitly to leave the sound alone. The plan was the game would play a sound *if* the discarded card *could* be in Bob's hand, and then all players have to check for themselves whether the card is actually on chop or not.

There is one way we could still make this feature request work. As of #1919 (also from yours truly), the game state *does* keep track of some minimal basic information about the player notes --- we needed this because we wanted our future efficiency calculation to include cards that were noted as gotten. It looks like I had the good foresight to include the "chop move" among that information, so we could make it work.

But with that said, I am not sure I would *want* to implement this, for the following reasons:

[...]

PR #2231, opened 2021-07-27
DDA fixup for #2222

vEnhance:
See discussion thread in the issue #2222

### DIFF
--- a/client/src/game/reducers/getSoundType.ts
+++ b/client/src/game/reducers/getSoundType.ts
@@ -99,7 +99,7 @@ export default function getSoundType(
         const previouslyDiscardedCard =
           originalState.deck[originalState.stats.doubleDiscard];
         if (
-          cardRules.canPossiblyBeFromClues(
+          cardRules.canPossiblyBeFromCluesOnly(
             discardedCard,
             previouslyDiscardedCard.suitIndex,
             previouslyDiscardedCard.rank,
--- a/client/src/game/reducers/getSoundType.ts
+++ b/client/src/game/reducers/getSoundType.ts
@@ -99,10 +99,10 @@ export default function getSoundType(
         const previouslyDiscardedCard =
           originalState.deck[originalState.stats.doubleDiscard];
         if (
-          discardedCard.possibleCardsFromClues.some(
-            ([suitIndex, rank]) =>
-              previouslyDiscardedCard.suitIndex === suitIndex &&
-              previouslyDiscardedCard.rank === rank,
+          cardRules.canPossiblyBeFromClues(
+            discardedCard,
+            previouslyDiscardedCard.suitIndex,
+            previouslyDiscardedCard.rank,
           )
         ) {
           // A player has discarded *in* a double discard situation
--- a/client/src/game/reducers/getSoundType.ts
+++ b/client/src/game/reducers/getSoundType.ts
@@ -90,28 +90,23 @@ export default function getSoundType(
         return SoundType.DiscardClued;
       }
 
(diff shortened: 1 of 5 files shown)


client/src/game/reducers/getSoundType.ts
```diff
@@ -90,28 +90,24 @@ export default function getSoundType(
...
-        couldBeLastDiscardedCard &&
         !metadata.hardVariant &&
         !handRules.isLocked(nextPlayerHand, currentState.deck)
       ) {
-        // A player has discarded *in* a double discard situation
-        return SoundType.DoubleDiscard;
+        const previouslyDiscardedCard =
+          originalState.deck[originalState.stats.doubleDiscard];
+        if (
+          cardRules.canPossiblyBe(
```
timotree3, line 102: Does canPossiblyBe use asymmetric information? I think it might have been intentional that the old code only considered explicit information from clues.

Zamiell:
```ts
export function canPossiblyBe(
  card: CardState,
  suitIndex: number | null,
  rank: number | null,
): boolean {
  if (suitIndex === null && rank === null) {
    // We have nothing to check
    return true;
  }
  return card.possibleCardsForEmpathy.some(
    ([s, r]) =>
      (suitIndex === null || suitIndex === s) && (rank === null || rank === r),
  );
}
```

Zamiell: its using empathy, thus it appears that it only uses symmetric information.

the fact that timo had to ask this question means that the function is incorrectly named, which is a bug.
can you rename it to something better?
maybe `canPossiblyBeFromEmpathy()`

vEnhance: `canPossiblyBe` a wrapper around `possiblyCardsForEmpathy`, so it's whatever the empathy shows.

We can change it to use `possibleCardsFromClues` instead but then I'd need to tweak the question mark code as well.

Zamiell:
```ts
  // possibleCardsForEmpathy is a one-dimensional array of tuples
  // It contains a tuple for each specific card that is still possible based on everything the
  // player holding it should know so far
  // Do not access this by the index; filter the array to find the remaining cards that you need
  readonly possibleCardsForEmpathy: ReadonlyArray<readonly [number, number]>;
```

ok so its based on assync info, thats bad

Zamiell: > We can change it to use possibleCardsFromClues instead but then I'd need to tweak the question mark code as well.

Yes, as I've said earlier, both the sounds and the question mark code should not use assync info

vEnhance: huh. didn't realize the empathy used async info, ok

Zamiell:
looks good now
does the question mark code also need to change?

vEnhance:
>  does the question mark code also need to change?

I guess I can put that in the same pull request yeah

I'll also make a new routine on cardRules so I can stop typing out this `some` idiom repeatedly and rename the old one.

vEnhance:
Quick thought:

1. Would it be useful to have an X mark over the question mark if asymmetric info rules out the card (just like pips and ranks get crossed out)
2. If so, should I put it in this PR or make a new one.

Zamiell:
i would prefer question marks to always show for simplicity.
its non-obvious why a question mark would be crossed out.

vEnhance:
OK. Then feel free to merge the PR as is.

Zamiell:
ok thanks

PR #2234, opened 2021-07-27
Fix shared replay controls

### LINKED ISSUE #2232: Replay controls get hidden in shared replay if keyboard is used to scroll to end
image

1. Open a shared replay
2. Click the arrow to go to the end
3. Press the left arrow key
4. Press the right arrow key

timotree3:
Fixes #2232


Zamiell:
thanks

PR #2253, opened 2021-09-04
Draw duplicate pips of a separate suits differently

### LINKED ISSUE #2251: Modify the second suit pip of double almost same suit variants
E.g. "Light Pink and Pink", or "Muddy Rainbow and Cocoa Rainbow" have this problem: the cards look very similar, and can be confused. One workaround is colorblind mode, because it always shows the letter in corner.

My proposal which works for non-colorblind mode too is to change the pip in some way to make it distinguishable form the base pip. Maybe add some pattern around the pip.

DarthGandalf:
Fix #2251


.cspell.json
```diff
@@ -141,6 +141,7 @@
         "kadda",
         "keldon",
         "keldon's",
+	"keyable",
```
Zamiell, line 144: looks like you manually inserted this instead of using the cspell took, because the whitespace is wrong

client/src/game/ui/drawPip.ts
```diff
@@ -3373,6 +3375,47 @@ export default function drawPip(
...
+      for (let i = 0; i < colors.length; ++i) {
+        grad.addColorStop(i / (colors.length - 1), colors[i]);
+      }
+      ctx.fillStyle = grad;
+    } else {
+      ctx.fillStyle = suit.fill;
+    }
+    // Workaround for Konva.Shape's Context not supporting evenodd
+    // This is used in e.g. Empathy view
+    interface Keyable {
```
Zamiell, line 3401: i think this is better written as
```ts
type Keyable = Record<string, any>
```
correct me if i am wrong

.cspell.json
DarthGandalf: I didn't know cspel can edit this file. Fixed.

client/src/game/ui/drawPip.ts
DarthGandalf: You're right. I just don't know any typescript

client/src/game/ui/drawPip.ts
```diff
@@ -3373,6 +3375,45 @@ export default function drawPip(
...
+      const colors = suit.fillColors;
+      const grad = ctx.createLinearGradient(0, -100, 30, 100);
+      for (let i = 0; i < colors.length; ++i) {
+        grad.addColorStop(i / (colors.length - 1), colors[i]);
+      }
+      ctx.fillStyle = grad;
+    } else {
+      ctx.fillStyle = suit.fill;
+    }
+    // Workaround for Konva.Shape's Context not supporting evenodd
```
DarthGandalf, line 3399: Konva fixed this in version 8, but this project still uses version 3.4.1

Zamiell:
cspell is still bugged, you added a newline at the end

DarthGandalf:
Fixed everything except for "unknown", it needs "any".

      TS2352: Conversion of type 'CanvasRenderingContext2D' to type 'Record<string, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Index signature is missing in type 'CanvasRenderingContext2D'.

Zamiell:
yeah, so cast it as unknown first exactly like the error message describes

DarthGandalf:
Right. I was trying to cast the wrong variable, so it didn't work

client/src/game/ui/drawPip.ts
```diff
@@ -3373,6 +3375,44 @@ export default function drawPip(
...
+    // This is used in e.g. Empathy view
+    type Keyable = Record<string, unknown>;
+    // eslint-disable-next-line
```
Zamiell, line 3402: why is this here?

DarthGandalf: 3402:50  error  ["_context"] is better written in dot notation                                                       @typescript-eslint/dot-notation

DarthGandalf: But the dot notation also shows error, because that type isn't supposed to have such field

DarthGandalf:
The whole Keyable is there only to workaround this issue with konva<8.0.0. Do you see some solution for this which doesn't involve disabling eslint?

Ideally, probably, update konva to >=8.0.0

Zamiell:
upgrading konva is complicated, when we tried last time there were some bugs, but if you take the commit where we upgraded and then fix the bugs from there, i would accept a PR for that

DarthGandalf:
Yeah, I'm not up to that task yet.

DarthGandalf:
Lint error is better than the compilation error, at least

Zamiell:
see my commit on how to model the data properly. its kind of a hack because the real type of the argument is the KonvaCanvasType thing but I don't know what it is

Zamiell:
thanks and nice work

DarthGandalf:
It looks like some internal Konva's type

Zamiell:
its some Konva type but presumably it is exposed for use in the typescript definitions

PR #2254, opened 2021-09-19
[RFC] Add 2 dockerfiles and a docker-compose file.

boweiliu:
This should simplify the installation instructions
and "how to run" instructions down to 1-3 lines:

```cp .env_docker_template .env && docker-compose up -d```

can result in an instantly running hanab live server
inside 2 docker containers (1 postgres, 1 go server).

The build process for these docker containers could
be further automated in a github actions script that runs

```docker-compose build && docker-compose push```

If there's any interest in this, I can help get this PR merged!

Otherwise I'm perfectly happy to keep it in a fork.

### CODE REVIEW

File: .env_docker_template [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+
+# HTTPS (TLS) Configuration
+# If blank, it will default to HTTP instead of using HTTPS
+TLS_CERT_FILE=
+TLS_KEY_FILE=
+
+# The PostgreSQL database configuration
+# If "DB_HOST" is blank, it will default to localhost
+# If "DB_PORT" is blank, it will default to 5432 (the default PostgreSQL port)
+DB_HOST=hanab-live-db
```
boweiliu, line 32: this is the only change from .env_template

Zamiell:
these files have text, so i think they should be spell checked
better to fix the actual spelling errors

boweiliu:
Thanks, can you rerun the actions now?

Zamiell:
why using node 12? i think the linter will fail on versions prior to 16

Zamiell:
also can you explain thought process behind separating server and db into separate docker images?

boweiliu:
> why using node 12? i think the linter will fail on versions prior to 16

i can try, was having some issues locally on 16 but i've also had some problems with other projects so might be my local setup.

> also can you explain thought process behind separating server and db into separate docker containers?

it's recommended as best practice in the industry, see 

https://docs.docker.com/config/containers/multi-service_container/

```
A container’s main running process is the ENTRYPOINT and/or CMD at the end of the Dockerfile. It is generally recommended that you separate areas of concern by using one service per container. That service may fork into multiple processes (for example, Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application. You can connect multiple containers using user-defined networks and shared volumes.



```

and

https://www.tutorialworks.com/containers-single-or-multiple-processes/#:~:text=One%20of%20these%20principles%20is,process%20running%20inside%20the%20container.

boweiliu:
Node 16 works for me, updated!

boweiliu:
Also see https://github.com/SashaIr/hanabi-live who had a similar idea (dockerfile + docker-compose).

Zamiell:
ok thanks

PR #2264, opened 2021-10-30
Add critical fours variants

### LINKED ISSUE #2168: Variant (Re)Proposal: Critical Fours

bhamrick:
Closes #2168. Tested a no variant game and a critical fours game to completion and they both appeared to work as expected.


bhamrick:
Hang on, changes don't seem to include the fix sorry.

bhamrick:
Ok, the changes _do_ include the fix. I was confused a minute ago because the implementation in client/src/game/rules/deck.ts doesn't match the implementation in server/src/game_deck.go (the ts version sets amountToAdd to 2 at the top of the function, while the go version leaves it at the default value of 0).

Zamiell:
these functions suck, can you refactor them so that they return values instead of use a variable that is mutated?

Zamiell:
that will prevent future bugs

bhamrick:
Sure.

bhamrick:
I quickly tested what I think is all the game types:

* Normal
* Reversed
* Up or Down
* One of Each (single and double dark)
* Crit Fours

and the decks had the expected distributions.

client/src/game/rules/deck.ts
```diff
@@ -30,33 +33,41 @@ export function numCopiesOfCard(
   rank: number,
   variant: Variant,
 ): number {
-  // In a normal suit of Hanabi,
-  // there are three 1's, two 2's, two 3's, two 4's, and one 5
-  let amountToAdd = 2;
+  // This implementation mirrors numCopiesOfCard in server/src/game_deck.go
```
Zamiell, line 36: good thank you

PR #2265, opened 2021-10-30
fixed pregame change options, tooltip repositioning

DrAnax:


DrAnax:
The "Change Options" button is also enabled and clickable for non-game-hosts, but the gray "Change" button does nothing (only the game host can change the options). Is this the intended behaviour or should I change something?

DrAnax:
Closes #2266

Zamiell:
> Is this the intended behaviour or should I change something?

um i think it should be clickable for everyone, but if non-leaders click it, it does a suggestion that goes into the chat that the leader can click or whatever

PR #2283, opened 2021-11-05
Clearer icon in history

### LINKED ISSUE #1908: Clearer options icon in games history
In "show history", if a game has only one additional option (e.g. speedrun) that icon should be displayed instead of a generic "+"

DrAnax:
Closes #1908


Zamiell:
good change

PR #2287, opened 2021-11-11
Change tooltip library

DrAnax:
All tooltips are created, updated and repositioned from a single module.

### DIFF
--- a/client/src/chat.ts
+++ b/client/src/chat.ts
@@ -590,7 +590,7 @@ export function add(data: ChatMessage, fast: boolean): void {
       match = regex.exec(line);
     }
   }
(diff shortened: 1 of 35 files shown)


Zamiell:
good.

you named the file konvaTooltipsts, but shouldn't it be named tooltips.ts?
that way, if you change the graphics library, then you wouldn't have to necessarily touch any other files

DrAnax:
the konvaTootps.ts deals with konva elements - nothing to do with the real tooltip library. It steals all mouse events and exposes them to the tooltip when appropriate

Zamiell:
ok, i see

PR #2289, opened 2021-11-29
Fix table creation commands

timotree3:
This PR disables exclamation mark validation when calling `isTableNameValid` from table_create because it was falsely invalidating any table name starting with an exclamation mark. The purpose of the validation is still fulfilled by code later in the function.


Zamiell:
thanks tim

Issue #1263, opened 2020-06-16
change replay leader should be an interactive tooltip where you click on the player to pass lead to instead of using an alert()


Zamiell:
@padiwik

padiwik:
So like the current alert but with buttons for each player instead of typing a number? You can't just do players since you want to be able to give lead to any spectator

Zamiell:
yeah i meant spectator

DrAnax:
Closed by #2290

PR #2290, opened 2021-11-29
Modals

DrAnax:
- All modals can be cancelled by clicking outside or by pressing escape
- Create game redone, the extra options are in two columns (unless screen is too narrow)
- Set turn, change leader, set efficiency modifier are now modals

Edit: and enter table password too

UI changes
- Set leader on mobile changes on double tap
- Set efficiency modifier changes on click (desktop) or double tap (mobile)


DrAnax:
Closes #1263

PR #2291, opened 2021-11-30
preselection in dialogs, enter for submit in create game

DrAnax:
- Set turn, play card and set modifier dialogs have the input value preselected
- Pressing ENTER on create game dialog submits the form


Zamiell:
thanks

PR #2301, opened 2021-12-06
Time display and calculations use absolute values

DrAnax:
Timer calculations now use absolute values given that in non timed games time is measured using negative values.


DrAnax:
Closes #2095

Zamiell:
good fix thank you

PR #2305, opened 2021-12-07
Fix morphed draggable

DrAnax:
Move blank card detection to keyboard.ts


timotree3:
I was hoping the keyboard controls would have parity with the misplay check and the unknown card morph prompt that the mouse controls have at https://github.com/Zamiell/hanabi-live/blob/b01f61984c01b3ddeab71f535654a540afdf88f5/client/src/game/ui/LayoutChild.ts#L140-L149

DrAnax:
> I was hoping the keyboard controls would have parity with the misplay check and the unknown card morph prompt that the mouse controls have at
> 
> https://github.com/Zamiell/hanabi-live/blob/b01f61984c01b3ddeab71f535654a540afdf88f5/client/src/game/ui/LayoutChild.ts#L140-L149

I don't know how to do this with a non-blocking dialog.

Issue #818, opened 2019-12-04
Feature request: hidden vote to terminate


aliblong:
I think it would be valuable to be able to express one's desire to terminate a game in such a way that it is not communicated to the team unless a supermajority (say >=60%) agrees.

Basically, each player would have a "vote to terminate" switch that they could toggle on and off as they wish. If a supermajority ever have the switch toggled on, the game is terminated. This means information about the game state is not leaked when a player states a desire to terminate. Furthermore, this empowers us to move to a paradigm where players will not unilaterally terminate games (the nuclear option) against the will of the team. The nuclear option can and probably should still exist, since ultimately, a game's continuance relies on omnilateral participation; however, I think it will become socially unacceptable to use for reasons other than an emergency or "restart because someone else just showed up and wants to play".

aliblong:
@jakestiles

jakestiles:
In favor. :) Sometimes the majority of players wish to terminate but there’s currently no way to tell unless someone wants to so much that they are willing to speak up about it.

Zamiell:
> click = VTK, possibly with timer, control + click = instant termination?

unfortunately, there is not room on the UI to add a separate button for VTK, so this suggestion seems reasonable to me.

timotree3:
I would suggest that clicking the skull triggers the vote, and adds a visual indicator to the skull (e.g. a bright outline), and clicking subsequently triggers instant termination (coupled with the confirmation message for players with <1000 games)

Zamiell:
agree with that

DrAnax:
> I would suggest that clicking the skull triggers the vote, and adds a visual indicator to the skull (e.g. a bright outline), and clicking subsequently triggers instant termination (coupled with the confirmation message for players with <1000 games)

That's tricky given that the dialog will be a modal and clicking outside will either cancel the dialog (if it's cancellable) or do nothing (if it isn't).

how about showing something like:

![image](https://user-images.githubusercontent.com/2622043/143892542-76831a23-dc70-4398-b3db-84fa0dd8a706.png)

Zamiell:
> That's tricky given that the dialog will be a modal

do we need to have a dialog for a VTK?
i dont think so.
by default, just have clicking it start a VTK.
reason: misclicking on the button in this case is far less destructive than a misclick causing a termination

DrAnax:
[...]

PR #2308, opened 2021-12-09
Vote to kill

DrAnax:
Some bits of democracy for Hanab players


client/src/game/rules/segment.ts
```diff
@@ -22,6 +22,7 @@ export function shouldStore(
     action.type === "gameOver" &&
     action.endCondition !== EndCondition.Timeout &&
     action.endCondition !== EndCondition.Terminated &&
+    action.endCondition !== EndCondition.VotedToKill &&
```
Zamiell, line 25: this has to be added at the end or else all the idle timeouts in the database will show as the wrong thing

server/src/command_table_vote_for_termination.go
```diff
@@ -0,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"strconv"
+)
+
+// commandTableTerminate is sent when the user clicks the terminate button in the bottom-left-hand
```
Zamiell, line 8: comment is inaccurate

client/src/game/ui/drawUI.ts
```diff
@@ -1053,7 +1077,7 @@ function drawScoreArea() {
     });
     terminateButton.tooltipName = "kill";
     terminateButton.tooltipContent =
-      "Terminate the game, ending it immediately.";
+      'Vote to terminate the game (click again to cancel the vote).<br><span style="padding-left: 30px;">Double click to terminate immediately.</span>';
```
Zamiell: i thought we wanted right click to terminate?

DrAnax: Indeed, but it felt to me more natural to have similar events for mouse and touch: click / tap, dbl click and dbl tap. I will branch this and make a new PR, accept either.

client/src/game/ui/drawUI.ts
```diff
@@ -1077,7 +1077,7 @@ function drawScoreArea() {
     });
     terminateButton.tooltipName = "kill";
     terminateButton.tooltipContent =
-      'Vote to terminate the game (click again to cancel the vote).<br><span style="padding-left: 30px;">Double click to terminate immediately.</span>';
+      'Vote to terminate the game (click again to cancel the vote).<br /><span style="padding-left: 30px;">Double click to terminate immediately.</span>';
```
DrAnax, line 1080: The `<br />` is a relic of xhtml and has been long abandonned.

https://www.google.com/search?q=%3Cbr%3E+vs+%3Cbr+%2F%3E&oq=%3Cbr%3E+vs+%3Cbr+%2F%3E

The span's role is to indent the second paragraph:

image

vs

image

I found it unecessary to to place the style in hanabi.css because it's a dynamically created tooltip.

DrAnax:
Closes #818

server/src/constants.go
```diff
@@ -39,6 +40,7 @@ const (
 	EndConditionStrikeout
 	EndConditionTimeout
 	EndConditionTerminated
+	EndConditionTerminatedByVote
```
Zamiell, line 43: this has to be the final thing or else it will fuck up the database, meaning that it will shift all the ids and then the existing ids will no longer match

DrAnax: makes sense

DrAnax: There should be in golang something similar to c# to prevent constants renumbering e.g

```
enum ErrorCode : ushort
{
    None = 0,
    Unknown = 1,
    ConnectionLost = 100,
    OutlierReading = 200
}
```

Zamiell: yes you can manually assign values in that way instead of using iota

PR #2319, opened 2021-12-14
Auto remove pipe during note edit

DrAnax:


Zamiell:
why does this PR include submodules relating to the server

DrAnax:
> why does this PR include submodules relating to the server

no idea, possibly me trying to setup dev tools properly yesterday, let me fix that

Zamiell:
rest in pieces, pipes

PR #2324, opened 2021-12-14
Revert 4775ab1a bug

DrAnax:


DrAnax:
Fixes wrong timer indication for first player only

PR #2326, opened 2021-12-14
Int not number

### LINKED ISSUE #1985: /random error message should say "integer" instead of "number"

DrAnax:
Closes #1985


Zamiell:
looks like this is part of some other PR? is that a mistake?

DrAnax:
Yes, my bad

PR #2327, opened 2021-12-14
Total games only

### LINKED ISSUE #2166: Games Played
I believe that in games played that kinda count the experience of the players, we shouldn't count the terminated and the bombed out.

DrAnax:
Closes #2166


Zamiell:
looks like this is part of some other PR

PR #2329, opened 2021-12-14
Cursor as class

### LINKED ISSUE #2263: Mouse cursor changing to image makes server requests
/client/src/game/ui/cursor.ts:32-57 is responsible for mouse cursor changes.

When the cursor changes to an image (look.png, edit.png) this makes each time a new server request for the image.

DrAnax:
Closes #2263

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -278,6 +278,7 @@
         "tablesorter",
         "tagdelete",
         "taglist",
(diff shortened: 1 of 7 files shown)


Zamiell:
it looks like this PR has stuff about tags in it, why is that

PR #2331, opened 2021-12-14
Set cursor via class, not style

DrAnax:


Zamiell:
much better, thank you

Zamiell:
does this also fix the bug where it constantly asks the server for the new icon image?

DrAnax:
Yes, that's the point

Zamiell:
ok, nice.
its also better in code because style should be in a css file lol

DrAnax:
Users need to reset cache or they won't see the changes

PR #2340, opened 2021-12-16
Create input and allow value encoding

DrAnax:
Allows users to write notes using quotes, less/greater than etc. (Nitrate)


Zamiell:
thanks

PR #2346, opened 2021-12-16
Special keys removing the pipe are ignored the first time

### LINKED ISSUE #2344: when making a note, backspace works incorrectly
step 1 - make the following note on a card:
[y2] p

step 2 - right click the card again to bring up the note box

step 3 - press backspace

expected: only the pipe and space will be deleted
observed: the "p" character is deleted

DrAnax:
Closes #2344


Zamiell:
doesn't this break the case of e.g. pressing home?

DrAnax:
> doesn't this break the case of e.g. pressing home?

It ignores all of [Arrows, Home/End, Del, Backspace]. Want only some of those?

Zamiell:
yes, home should go home.
the rest are fine as being a no-op.

Zamiell:
probably

Zamiell:
ok, lets try this out for a few days ><

Issue #2323, opened 2021-12-14
How to deal with max number of players changing in pregame?


DrAnax:
- Can table owner change the number of players in pregame?
- Can guests propose a set of options including max number of players

If yes in either case:
- What if the new number of players is less than the present number of players?

Zamiell:
> Can table owner change the number of players in pregame?

i guess

> Can guests propose a set of options including max number of players

i guess

> What if the new number of players is less than the present number of players?

boot the final player(s)

DrAnax:
> boot the final player(s)

> don't allow changing the number of players in this case OR
> default to the current number of players

I need a final decision by tomorrow

Zamiell:
what i said makes more sense because the table leader has the authority to boot people from the table with /kick anyway

mangopie1:
Make sense. Otherwise owner have to manually kick them before making the change. And if someone join again in the meantime, owner have to kick again. So auto kick sounds better UX

Zamiell:
fixed now

PR #2348, opened 2021-12-17
Max players change option

DrAnax:
- Max players is now saved in local storage
- During pregame, host can change the max number of players
- During pregame, guests can propose new max players (warning if this is less than the number present ones)
- If new value of Max Players is less than the number present players, extra players are kicked with a message.

### DIFF
--- a/client/src/lobby/createGame.ts
+++ b/client/src/lobby/createGame.ts
@@ -355,6 +355,7 @@ function submit() {
       tableID: globals.tableID,
       name,
       options,
+      maxPlayers,
     });
   }
 
@@ -367,11 +368,13 @@ function submit() {
 
 function acceptOptionsFromGuest(options: Options) {
   const name = options.tableName;
+  const maxPlayers = options.maxPlayers;
 
   globals.conn!.send("tableUpdate", {
     tableID: globals.tableID,
     name,
     options,
+    maxPlayers,
   });
 }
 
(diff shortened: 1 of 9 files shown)


Zamiell:
why is max players stored in localstorage instead of in the db like all of the other create-table-related options are?

DrAnax:
Closes #2323 

> why is max players stored in localstorage instead of in the db like all of the other create-table-related options are?

That's what happens with other table options. The local storage values are used in the UI for new table dialog

Zamiell:
> That's what happens with other table options. The local storage values are used in the UI for new table dialog

shouldn't we store it in the database so that every table option works in an identical way?

DrAnax:
> shouldn't we store it in the database so that every table option works in an identical way?

Do you really need that option in the database? The table could have started with more max than the present anyway. I don't think that variable is relevant to the table or the game.

Zamiell:
Alice creates Hanabi games every day.
She only plays timed games of the rainbow variant and with a max players of 4.
Most of the time, she is on her work computer.
The website handily remembers all of these options for her, so that she doesn't have to reselect anything.

Alice gets a new computer, and sees that the website handily remembers all of her settings.
She doesn't notice right away that unlike all of her other settings, the max players setting was not remembered, and defaults to 5.
Alice creates a table like she normally does, but 5 people end up joining the game.
Alice is very frustrated with the website - why did it remember all of her other options but not the max players option?

DrAnax:
> Alice is very frustrated with the website - why did it remember all of her other options but not the max players option?

Because she installed Windows? (omw)

DrAnax:
**Database upgrade**

PR #2352, opened 2021-12-17
Better history table

DrAnax:
Initial effort


server/src/http_shared_missing_scores.go
```diff
@@ -65,13 +65,14 @@ func httpSharedMissingScores(c *gin.Context) {
 		}
 	}
 
+	lastIndex := len(playerNames) - 1
 	data := &TemplateData{ // nolint: exhaustivestruct
 		Title:               "Missing Scores",
-		NamesTitle:          "Missing Scores for [" + strings.Join(playerNames, ", ") + "]",
+		NamesTitle:          "Missing Scores for " + strings.Join(playerNames[:lastIndex], ", ") + " and " + playerNames[lastIndex],
```
Zamiell, line 71: isnt this bugged when there is only one player

server/src/http_history.go
```diff
@@ -63,5 +63,5 @@ func httpHistory(c *gin.Context) {
 	if len(playerNames) > 1 {
 		data.NamesTitle = "Game History for [" + strings.Join(playerNames, ", ") + "]"
 	}
-	httpServeTemplate(w, data, "profile", "history")
+	httpServeTemplate(w, data, "profile-new", "history")
```
Zamiell, line 66: why are we using profile-new instead of profile

DrAnax: I was going back and forth between profile and profile-new. In the end I managed to make all the pages as I liked, but I forgot to transfer back.

server/src/http_shared_missing_scores.go
DrAnax: image

Zamiell:
you need to rebase this on main because i am seeing unrelated changes to this PR

server/src/http_shared_missing_scores.go
Zamiell: ok, nevermind then

PR #2359, opened 2021-12-18
Logger package

DrAnax:
Moved global logger into a separate package


server/src/game_player.go
```diff
@@ -189,15 +191,14 @@ func (p *GamePlayer) PlayCard(c *Card) {
 	// Update the progress
 	progressFloat := float64(g.Score) / float64(g.MaxScore) * 100 // In percent
 	progress := int(math.Round(progressFloat))
-	oldProgress := t.Progress
```
Zamiell, line 192: what does this have to do with logging?

DrAnax: That's a fix from linter warnings (ifshort)

server/src/http_google_analytics.go
```diff
@@ -40,7 +40,7 @@ func httpGoogleAnalytics(c *gin.Context) {
 	if cookie, err := r.Cookie("_ga"); err != nil {
 		// They don't have a cookie set, so set a new one
 		clientID = uuid.NewV4().String()
-		http.SetCookie(w, &http.Cookie{ // nolint: exhaustivestruct
+		http.SetCookie(c.Writer, &http.Cookie{ // nolint: exhaustivestruct
```
Zamiell, line 43: looks like there are many changes in this PR that have nothing to do with refactoring a logger package

DrAnax: Same as above

Zamiell: instead of changing code, just set the linter version to one from 2019

PR #2365, opened 2021-12-18
Take into account spaces before "(" and after ")"

DrAnax:


Zamiell:
should we write some tests?

DrAnax:
Possibly

DrAnax:
Ah, I need to push the test in a separate PR first, right?

Zamiell:
you can just do it in this one to make it easy

Zamiell:
thanks

Zamiell:
those are some good tests :)

PR #2369, opened 2021-12-20
Fix create game dialog

### LINKED ISSUE #2349: if max number of players has an invalid value, it should change to the default when focus leaves the box
with respect to the input box in the create game modal

DrAnax:
Closes #2349


Zamiell:
thanks

PR #2388, opened 2021-12-24
Show empathy on hidden cards during hypo

### LINKED ISSUE #2387: hypo improvements for starting player
after the last PR, the starting player's hand is hidden

however, if this hand has been previously clued, it should appear partially clued to every player in the shared replay, as it would appear to the player at the time of the hypothetical

right now, it does not do this, and is just completely hidden with a yellow border to indicate that it is clued

DrAnax:
Closes #2387


Zamiell:
i tried this out and exiting hypothetical doesn't seem to be resetting the cards back to normal.
is that because you aren't using the reducers?

DrAnax:
> i tried this out and exiting hypothetical doesn't seem to be resetting the cards back to normal. is that because you aren't using the reducers?

Let me check

Zamiell:
looks good!

Zamiell:
nice work

PR #2390, opened 2021-12-24
Hypo on keldon

### LINKED ISSUE #2389: when in hypo, cards can have wrong orientation
steps to reproduce:

1) turn on keldon mode
2) start a hypothetical on the turn of someone with cards that do not have a horizontal orientation
3) click and release a card

normally, this would reorient the card back to normal, but since the X/Y positions didn't change, no tween applies

DrAnax:
Closes #2389

### DIFF
--- a/client/src/game/ui/HanabiCardMouse.ts
+++ b/client/src/game/ui/HanabiCardMouse.ts
@@ -142,6 +142,7 @@ function mouseUp(this: HanabiCard) {
     return;
   }
 
+  // During replay hypo, skip the starting player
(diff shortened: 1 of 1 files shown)


Zamiell:
this pr is a little confusing, as bug #2389 doesn't have anything to do with empathy, right?

DrAnax:
> this pr is a little confusing, as bug #2389 doesn't have anything to do with empathy, right?

Actually, it did. Checking for empathy should occur in most possible mouse/touch actions

Zamiell:
ok

PR #2391, opened 2021-12-24
Correctly identify card for empathy cancel

DrAnax:
Unrelated to latest issue, but still a bug, fixed now.

HanabiCard.id (inherited from Konva.Group) returns a function, not a value.


Zamiell:
thanks

PR #2412, opened 2022-01-04
Rename Ruby to Orange

### LINKED ISSUE #2407: Variant: Very Ambigious (6 Suits)
Now that the reds are perfectly distinctable maybe rename Ruby to Orange to match the color

DrAnax:
Closes #2407

### DIFF
--- a/data/suits.json
+++ b/data/suits.json
@@ -301,11 +301,11 @@
     "pip": "hanabiGreen"
   },
   {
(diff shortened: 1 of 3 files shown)


Zamiell:
can we come up with a different name?
Orange is supposed to be a dual color suit

DrAnax:
Yes, will think about it and continue the PR

PR #2418, opened 2022-01-05
Vtk messages

### LINKED ISSUE #2393: Change end-of-game text
Is even worse than before now, before the vtk someone wrote vtk on chat and when 50% wrote it as well someone killed the game, now the game is suddenly killed and you have no idea that people where even voting. 

I think when the first one presses the button we should at least know that by changing the color of the skull

### LINKED ISSUE #2414: in a 2-player game, the termination message is incorrect
In a 2-player game, if Alice votes to kill the game, then the message that displays is "The players voted to terminate the game" instead of "Alice voted to terminate the game".

DrAnax:
Closes #2393 

Closes #2414

### DIFF
--- a/client/src/game/reducers/gameStateReducer.ts
+++ b/client/src/game/reducers/gameStateReducer.ts
@@ -152,6 +152,7 @@ function gameStateReducerFunction(
         action.playerIndex,
         state.score,
         metadata,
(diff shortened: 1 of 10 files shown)


Zamiell:
nice work

PR #2419, opened 2022-01-05
Update options using seed

### LINKED ISSUE #2417: Change options into a specific seed game
When I tried to change the option "deck blind play" for the a gauntlet seed I got this warning 
image
Is this the expected behavior ?

DrAnax:
Closes #2417


server/src/misc.go
```diff
@@ -423,3 +423,102 @@ func yesNoFromBoolean(option bool) string {
 	}
 	return "No"
 }
+
+func hasTableNameValidCommand(s *Session, d *CommandData, data *SpecialGameData) bool {
+	if strings.HasPrefix(d.Name, "!") {
```
Zamiell, line 428: return early

DrAnax: Not sure I understand your point

Zamiell: the way you wrote this function is an anti-pattern.
invert the logic in the if statement and return, then outdent the rest of the function

DrAnax: I didn't write any function, I just enclosed existing code within a function block, since it is used identically in two different places (table_create, table_restart). Nevertheless I'll redo

Zamiell: right - it couldn't be written using return early pattern in the place it was before because there was logic following the if statement (which is no longer the case)

Zamiell:
better, thanks

PR #2427, opened 2022-01-06
fix regexp

### LINKED ISSUE #2059: discord roles dont get chat filled properly

DrAnax:
Closes #2059 

Cannot test (but [code](https://github.com/Hanabi-Live/hanabi-live/blob/bcbe24fe6ef18f1591ffd9aaab68980e3f7bed5b/server/src/chat.go#L93) now fetches the correct user id)

(based on issue screenshot)

### DIFF
--- a/server/src/chat.go
+++ b/server/src/chat.go
@@ -19,7 +19,7 @@ const (
 )
 
 var (
-	mentionRegExp = regexp.MustCompile(`&lt;@&!*(\d+?)&gt;`)
+	mentionRegExp = regexp.MustCompile(`&lt;@&*!*(\d+?)&gt;`)
 	channelRegExp = regexp.MustCompile(`&lt;#(\d+?)&gt;`)
 )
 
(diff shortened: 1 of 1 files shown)


DrAnax:
Rerun the tests

Zamiell:
i think the `&` should have a `*` after it right?
since it isn't in every single mention

DrAnax:
> i think the `&` should have a `*` after it right? since it isn't in every single mention

According to the screenshot, there's only one `&`. Console tests show that the user_id is grabbed correctly.

Zamiell:
correct me if I am wrong, but I think this regex is used to parse both user mentions and role mentions, which are distinct from each other.
the former does not have an ampersand in it, right?

DrAnax:
Right:

```
mentionRegExp = regexp.MustCompile(`&lt;@&*!*(\d+?)&gt;`)
test &lt;@&921716503051632650&gt;
Discord ID: 921716503051632650

test &lt;@921716503051632650&gt;
Discord ID: 921716503051632650
```

vs

```
mentionRegExp = regexp.MustCompile(`&lt;@&!*(\d+?)&gt;`)
test &lt;@&921716503051632650&gt;
Discord ID: 921716503051632650

test &lt;@921716503051632650&gt;
Discord ID: not found
```

server/src/chat.go
```diff
@@ -79,9 +79,9 @@ func chatServerSendPM(s *Session, msg string, room string) {
 }
 
 func chatFillMentions(msg string) string {
-	if discord == nil {
-		return msg
-	}
+	// if discord == nil {
```
Zamiell, line 82: what

DrAnax: oops, debug, sorry

Zamiell:
good fix, thank you

PR #2428, opened 2022-01-06
No warning for empty clue in speedrun

### LINKED ISSUE #2198: The empty clues warning during speedruns.
Hello. An issue I have found is that sometimes, when you pre-clue a card that gets discarded, the clue will touch nothing and there will be a warning that pops up, saying that you cannot give a clue that touches no cards. Except, that basically kills the run because of having to close the warning. Could this be changed to more of a sound effect?

DrAnax:
Closes #2198


Zamiell:
good fix, thanks

PR #2429, opened 2022-01-06
Skip over d/a/r/k

### LINKED ISSUE #2046: when assigning letters for suit abbreviations, it should skip over d, a, r, k
https://github.com/Zamiell/hanabi-live/blob/master/docs/FEATURES.md#card-identity-notes

according to that 

[any dark suit] | [the same as the "normal" version of the suit]
-- | --

so pink should be i and prism should be p (probably) 
I found out that d is working but I dont know why

DrAnax:
Closes #2046


Zamiell:
good fix, thank you

PR #2437, opened 2022-01-06
Update repo

DrAnax:


DrAnax:
Can actions somehow check which part of the code the PR changes and only run those specific tests?

Zamiell:
i believe so

Zamiell:
you'll have to do more research into that, i'm not a github actions expert

DrAnax:
> you'll have to do more research into that, i'm not a github actions expert

https://stackoverflow.com/a/63822945/180243

DrAnax:
Also: https://github.com/tj-actions/changed-files

Zamiell:
ah good, that seems simple enough

Zamiell:
submit a PR that adds those fields to the ci.yml file for client and server respectively

PR #2439, opened 2022-01-06
Replace bool with strings

DrAnax:


Zamiell:
can you explain the rational behind this?
it seems like an objectively worse way to structure the code, since people could just mimic that string to maliciously spam the "website-development" channel, for instance

DrAnax:
Haven't thought about the malicious intend - my bad

The code block in question was:

```
// Some messages are also sent to website-development
if sendMessageToWebDevChannel {
	discordSend(discordChannelWebsiteDev, d.Username, rawMsg)
	sendMessageToWebDevChannel = false
}
```

Zamiell:
on the face of it, the code you pasted doesn't seem like it would affect the connection lock, right?

DrAnax:
I felt that in a non-linear environment that's the only potential problem. Nothing else in my commits looks remotely suspicious imo.

PR #2441, opened 2022-01-07
Update spinner to reflect the updated cards

calvinballing:
Updates the spinner to use the Hanab blue-card symbol instead of the older diamond symbol.  Also changes the rotation style to be more three dimensional, and displays the back of the card when facing away.

New:
ajax
 
Old:
ajax-old


Zamiell:
neat

Zamiell:
nice work

Zamiell:
can you just get rid of the unused ajax image.
we dont need to keep unused images in there, we can always get the old version of the file from git

Zamiell:
also are you on discord?

calvinballing:
> can you just get rid of the unused ajax image. we dont need to keep unused images in there, we can always get the old version of the file from git if we need it

I've removed the old one in this update.  I just joined the Discord, and I'm Jim#4145.

DrAnax (approved):

Zamiell:
Ok, thanks. I was just curious if u had another ID, since I didn't recognize your name

PR #2445, opened 2022-01-07
Shows uptime before shutting down

### LINKED ISSUE #2440: server shutdown message that gets put into discord should also say how long it was online for

DrAnax:
Closes #2440


Zamiell:
thank you

PR #2465, opened 2022-01-10
Fix client jest config

DrAnax:


DrAnax:
Tests are again green in VSC

PR #2470, opened 2022-01-11
Draw special 1s and 5s according to their special rank property

DarthGandalf:
hanabi-brown5
hanabi-lpink1
hanabi-muddy1
hanabi-null1
hanabi-pink5


Zamiell:
nice work

PR #2472, opened 2022-01-11
Fixing Dockerfile after latest refactoring

N1trate:
I don't know much about node/go, but that looks right and got it to work on my dev.


Zamiell:
i dont use docker but i'll take your word for it

PR #2473, opened 2022-01-11
Draw deceptive 1s, 5s

DarthGandalf:
hanabi-decep
hanabi-decep-5
hanabi-decep-cb
hanabi-decep-cb5
hanabi-decep-style
hanabi-decep-style5

### DIFF
--- a/packages/client/src/game/ui/drawCards.ts
+++ b/packages/client/src/game/ui/drawCards.ts
@@ -39,7 +39,9 @@ export default function drawCards(
   const suits = variant.suits.concat(unknownSuit);
   const pipTypes = new Set<string>();
 
+  let suitIndex = -1;
   for (const suit of suits) {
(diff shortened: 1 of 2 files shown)


Zamiell:
neat
i like it

Zamiell:
now fix your vscode so that you dont push code violating the linter to github

DarthGandalf:
I don't even use vscode

And that's why there is a CI

Zamiell:
what do you use, vim?

Zamiell:
> And that's why there is a CI

it spams my inbox if you keep pushing commits for useless linting fixes, so i would appreciate you getting it right before you push

DarthGandalf:
Yes, vim. I haven't configured automatic linter for js

Zamiell:
vim can lint in the same way vscode can AFAIK

PR #2476, opened 2022-01-11
Hypo to clipboard

### LINKED ISSUE #2466: "create a hypo to json exporter"
^ title are the words that Zamiel used to describe what I believe was being discussed on Discord today.

I can try and describe the process I'm envisioning:

- Play through a hypo of a certain seed (on the website)
- Somehow create a permanent record of the hypo; Zam mentioned it may be possible to make it exportable it as a .json, which then could be shared + viewed by others.

And for those curious, the larger aim of this would be to aid in a large-scale data collection effort

DrAnax:
Closes #2466

### DIFF
--- a/packages/client/src/lobby/createReplayJSON.ts
+++ b/packages/client/src/lobby/createReplayJSON.ts
@@ -8,13 +8,21 @@ import ReplayState from "../game/types/ReplayState";
 import globals from "../game/ui/globals";
 
 export default function createJSONFromReplay(room: string) {
-  if (globals === null || globals.state.replay.hypothetical === null) {
+  if (
+    globals === null ||
+    globals.store === null ||
+    !globals.state.finished ||
+    globals.state.replay === null ||
+    globals.state.replay.hypothetical === null
+  ) {
     chat.addSelf(
       '<span class="red">Error!</span>: You can only use that command during the review of a hypo.',
       room,
     );
     return;
   }
+  console.log("---------GLOBALS-----------");
+  console.log(globals);
 
   // Anonymize the players
   const fakePlayers = ["Alice", "Bob", "Cathy", "Donald", "Emily", "Frank"];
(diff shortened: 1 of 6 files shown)


packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -0,0 +1,226 @@
...
+    globals.state.replay.hypothetical === null
+  ) {
+    chat.addSelf(
+      '<span class="red">Error:</span> You can only use that command during the review of a hypothetical.',
+      room,
+    );
+    return;
+  }
+  // Anonymize the players
+  const fakePlayers = ["Alice", "Bob", "Cathy", "Donald", "Emily", "Frank"];
```
Zamiell, line 26: shouldn't this be a constant somewhere, probably in @hanabi/data

DrAnax: That's a good idea (btw is 'Frank' correct? I kinda guessed)

Zamiell: https://github.com/Hanabi-Live/hanabi-live/commit/d443c1f717f5a282d02154bde231b3f7d0f63e80

rogers-j:
Remove the extra `p` in 'clipboard'

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -22,10 +23,16 @@ export default function createJSONFromReplay(room: string) {
     return;
   }
 
+  let playerNames: string[] = [];
+  if (HYPO_PLAYER_NAMES.length >= globals.metadata.playerNames.length) {
```
Zamiell, line 27: is the if statement necessary?

DrAnax: Yes because the linter complains

Zamiell: with what error

DrAnax: It doesn't anymore :S

Zamiell:
nice work! this is a neat feature

PR #2481, opened 2022-01-12
Fix bug

DrAnax:


Zamiell:
SLR

Zamiell:
meaning break this up into multiple lines because that is too much for one line to do

DrAnax:
> 

Read it as "Sorry late reply", so I thought you were busy playing and went to sleep.

Zamiell:
no, SLR is my favorite acronym

https://midu.dev/single-line-responsability-haz-una-cosa-por-linea/

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -207,7 +207,8 @@ function getCardFromHypoState(
   if (globals.state.replay.hypothetical === null) {
     return 0;
   }
-  return globals.state.replay.hypothetical.states[stateIndex].hands[
+  const minStateIndex = Math.max(stateIndex, 0);
+  return globals.state.replay.hypothetical.states[minStateIndex].hands[
```
Zamiell, line 211: we can do better than this.
the line is still doing more than one thing.
keep breaking it up with good variable names that describe what is actually happening

DrAnax:
Why tests do not run?

Zamiell:
you can do even more SLR here

Zamiell:
how about this:
```ts
const stateOnHypoTurn = globals.state.replay.hypothetical.states[minStateIndex]; // is this the right variable name? what specific state is this?
const playerHand = state.hands[playerIndex];
return playerHand [slotIndex];
```

Zamiell:
much better thanks :)

PR #2483, opened 2022-01-13
Seed names

### LINKED ISSUE #2173: [bug] Accepted character range for seed names
The server communicates that only [`-A-Za-z0-9`] are valid characters for seed names, but this replay shows that it was possible to use a `#` character in the name. And it causes some wonky stuff when looking at other scores, for example.

https://hanab.live/replay/507894

DrAnax:
Closes #2173

### DIFF
--- a/server/src/command_replay_create.go
+++ b/server/src/command_replay_create.go
@@ -45,7 +45,12 @@ func commandReplayCreate(ctx context.Context, s *Session, d *CommandData) {
 	} else if d.Source == "json" {
 		// Before creating a new game and emulating the actions,
 		// ensure that the submitted JSON does not have any obvious errors
-		if valid, _ := isJSONValid(s, d); !valid {
+
+		if d.GameJSON == nil {
+			s.Warning("You must send the game specification in the \"gameJSON\" field.")
+			return
+		} else if valid, message := isJSONValid(s, d); !valid {
+			s.Warning(message)
 			return
 		}
 	}
--- a/server/src/command_replay_create.go
+++ b/server/src/command_replay_create.go
@@ -356,6 +356,7 @@ func isJSONValid(s *Session, d *CommandData) (bool, string) {
 	// Validate the characters
 	if d.GameJSON.Options.DetrimentalCharacters != nil &&
 		len(d.GameJSON.Characters) != len(d.GameJSON.Players) {
+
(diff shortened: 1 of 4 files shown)


DrAnax:
Optionally:

Fix existing seeds (search seeds containing `#`, `_`, `;`):

1. Test
  `SELECT id, seed, REGEXP_REPLACE(seed, '[#_;]', '-', 'g') AS becomes FROM "games" WHERE seed ~ '[#_;]';`

2. Change
  `UPDATE "games" SET seed = REGEXP_REPLACE(seed, '[#_;]', '-', 'g') WHERE seed ~ '[#_;]';`

3. Confirm
  `SELECT * FROM "games" WHERE seed ~ '[#_;]';`

server/src/misc.go
```diff
@@ -306,16 +306,13 @@ func toSnakeCase(str string) string {
...
-func fixTableName(name string) string {
+func truncateTrimCheckEmpty(name string) string {
 	// Truncate long table names
 	// (we do this first to prevent wasting CPU cycles on validating extremely long table names)
 	if len(name) > MaxGameNameLength {
 		name = name[0 : MaxGameNameLength-1]
 	}
 
-	// Remove any non-printable characters, if any
-	name = removeNonPrintableCharacters(name)
```
Zamiell, line 317: why

DrAnax: removeNonPrintableCharacters was obsolete; valid table/seed name is handled by regex

server/src/command_replay_create.go
```diff
@@ -45,7 +45,12 @@ func commandReplayCreate(ctx context.Context, s *Session, d *CommandData) {
 	} else if d.Source == "json" {
 		// Before creating a new game and emulating the actions,
 		// ensure that the submitted JSON does not have any obvious errors
-		if valid, _ := isJSONValid(s, d); !valid {
+
+		if d.GameJSON == nil {
```
Zamiell, line 49: this should be part of the isJSONValid function

DrAnax: it will break command_table_create.go:L91:96

Zamiell: oh I see

Zamiell:
great work, thanks

PR #2495, opened 2022-01-15
Darken the dark suit special 1/5

DarthGandalf:
Fix certain combinations: in "Brown-1s & White" w1 is null, but before
it was drawn as brown, which is technically correct but misleading.


Zamiell:
thanks

PR #2497, opened 2022-01-17
Removing Hmm from the emotes so we can use it normally

N1trate:
Just been told by Helana that I'm the only one that she's seen use that emote, and I'm using it against my will.
I'm happy there are no chatlog database table, but I wished we could check how often other people are really using it.

I can open an issue if we want to discuss it, or we can discuss it here.


timotree3:
I don't think usage stats are very helpful since every time I've seen someone use it, it has been accidental.

timotree3:
I'm down to remove the Hmm emoji though. Seems to communicate a lot less than Hmm does

timotree3 (approved):

PR #2506, opened 2022-01-18
Fix discord replace role on chat

DrAnax:


Zamiell:
the & is optional

DrAnax:
> the & is optional

`discordID` is found from the message string as the first match of the regex defined in chat.go:L24 as

```roleRegExp    = regexp.MustCompile(`&lt;@&(\d{17,19})&gt;`)```

According to that line, the `&` is not optional. That regex was found [here](https://github.com/discordjs/discord.js/blob/988a51b7641f8b33cc9387664605ddc02134859d/src/structures/MessageMentions.js#L227)

Zamiell:
ok, if you are sure

PR #2507, opened 2022-01-19
Change morphing modal:

DarthGandalf:
* Fix clicking
* Require only 1 click instead of 2
* Select the current card by default


Zamiell:
window.globals is never supposed to be used, can you refactor this to use the variables properly

DarthGandalf:
Done. I was surprised it used global vars like that, but had decided not to touch it

Zamiell:
ok, thanks

PR #2514, opened 2022-01-22
Properly search for discord roles in chat messages

DrAnax:


Zamiell:
thank you very much sir

PR #2518, opened 2022-01-24
Change password

### LINKED ISSUE #2511: Allow changing of passwords
Currently the only way to change a password is via the password reset, which requires providing of personal information and an admin to approve it, which is logical for actual resets, but not if the goal is to simply change the password while still having access to the old password.

I would submit a PR myself, if it weren't for the fact, that I have little to no experience with neither TS nor Go, so I'm hoping that someone with more experience would implement this.

What I have in mind is basically the password changing form of other sites, providing the old password and the new password.

DrAnax:
Closes #2511

### DIFF
--- a/server/src/http_login.go
+++ b/server/src/http_login.go
@@ -24,6 +24,7 @@ type HTTPLoginData struct {
 	IP                 string
 	Username           string
 	Password           string
+	NewPassword        string
 	NormalizedUsername string
 }
 
@@ -130,6 +131,35 @@ func httpLogin(c *gin.Context) {
 			} else if !match {
 				http.Error(w, "That is not the correct password.", http.StatusUnauthorized)
 				return
+			}
+
+			// If they have provided a new password, replace the old one
+			if data.NewPassword != "" {
+				var passwordHash string
+				if v, err := argon2id.CreateHash(data.NewPassword, argon2id.DefaultParams); err != nil {
+					logger.Error("Failed to create a hash from the submitted password for " +
+						"\"" + data.Username + "\": " + err.Error())
+					http.Error(
+						w,
+						http.StatusText(http.StatusInternalServerError),
+						http.StatusInternalServerError,
+					)
+					return
+				} else {
+					passwordHash = v
+				}
+				// Update the user
+				if err := models.Users.ChangePassword(user.ID, passwordHash); err != nil {
+					logger.Error("Failed to update user \"" + data.Username + "\" password: " + err.Error())
+					http.Error(
+						w,
+						http.StatusText(http.StatusInternalServerError),
+						http.StatusInternalServerError,
+					)
+					return
+				} else {
+					logger.Info("Password update for user \"" + data.Username + "\"")
+				}
 			}
 		}
 	} else {
@@ -286,6 +316,7 @@ func httpLoginValidate(c *gin.Context) (*HTTPLoginData, bool) {
 		)
 		return nil, false
 	}
+	newPassword := c.PostForm("newPassword")
 	version := c.PostForm("version")
 	if version == "" {
 		logger.Info("User from IP \"" + ip + "\" tried to log in, " +
@@ -445,6 +476,7 @@ func httpLoginValidate(c *gin.Context) (*HTTPLoginData, bool) {
 		IP:                 ip,
 		Username:           username,
 		Password:           password,
+		NewPassword:        newPassword,
 		NormalizedUsername: normalizedUsername,
 	}
 	return data, true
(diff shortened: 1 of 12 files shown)


Zamiell:
can you put any prettier related changes into a separate pr

server/src/http_login.go
```diff
@@ -131,6 +132,35 @@ func httpLogin(c *gin.Context) {
 				http.Error(w, "That is not the correct password.", http.StatusUnauthorized)
 				return
 			}
+
+			// If they have provided a new password, replace the old one
+			if data.NewPassword != "" {
```
timotree3, line 137: If NewPassword is not passed, does it default to `""` or `nil`?

DrAnax: It is empty string

server/src/http_login.go
```diff
@@ -131,6 +132,35 @@ func httpLogin(c *gin.Context) {
...
+						"\"" + data.Username + "\": " + err.Error())
+					http.Error(
+						w,
+						http.StatusText(http.StatusInternalServerError),
+						http.StatusInternalServerError,
+					)
+					return
+				} else {
+					passwordHash = v
+				}
```
timotree3, line 150: Isn't it cleaner not to introduce a mutable variable?
```suggestion
				passwordHash, err := argon2id.CreateHash(data.NewPassword, argon2id.DefaultParams)
				if err != nil {
					logger.Error("Failed to create a hash from the submitted password for " +
						"\"" + data.Username + "\": " + err.Error())
					http.Error(
						w,
						http.StatusText(http.StatusInternalServerError),
						http.StatusInternalServerError,
					)
					return
				}
```

Zamiell: no, because now you are scoping err inproperly

Zamiell: the rest of the function should never have access to err

server/src/http_login.go
```diff
@@ -131,6 +132,35 @@ func httpLogin(c *gin.Context) {
 				http.Error(w, "That is not the correct password.", http.StatusUnauthorized)
 				return
 			}
+
+			// If they have provided a new password, replace the old one
```
timotree3, line 136: Is there a reason why this code should be inside the `else` from `if user.OldPasswordHash.Valid {`? It seems like it's indepedent of whether they had an old hash or not

server/src/http_login.go
```diff
@@ -131,6 +132,35 @@ func httpLogin(c *gin.Context) {
...
+						http.StatusInternalServerError,
+					)
+					return
+				} else {
+					logger.Info("Password update for user \"" + data.Username + "\"")
+				}
+			}
 		}
 	} else {
 		// Check to see if any other users have a normalized version of this username
```
timotree3, line 166: Should we be erroring here if `data.NewPassword != ""`?

server/src/http_login.go
DrAnax: Not really, I can rewrite (and also make the password hashing an independent function)

server/src/http_login.go
timotree3: My point is that if someone is making a change password request, and the account doesn't exist, it seems apt to have the request fail instead of create a new account with the specified username + current password.

DrAnax: That part of the code never runs if this is a new user

DrAnax:
Shall I add now the missing semicolons in login.ts?

Zamiell:
if login.ts has missing semicolons, then why does the repo currently show as passing ci?
we need to fix that first if it is broken

DrAnax:
> if login.ts has missing semicolons, then why does the repo currently show as green / passing ci? we need to fix that first if it is broken

It shows failing on my screen and the PR has a red X at the end

image

Zamiell:
ok so the missing semicolons are only part of your changes then? i dont understand the question then

Zamiell:
when i said
"can you put any prettier related changes into a separate pr"
what I meant was:
"this pr should only have changes regarding to change password stuff, and not random code changes elsewhere in the code base"

DrAnax:
Ok, team resync-ed :)

Zamiell:
tim should I merge this?

server/src/http_login.go
timotree3: I know. I'm posting a comment on the `else` for `if exists {`. If `!exists && data.newPassword != ""`, for me that is alarming and we should error. Right now the code will proceed to create a new user as though it wasn't a "change password" request. It's not a huge deal since it seems like this can never happen in the current UI, but I just want to make sure I'm actually understood

Zamiell:
this PR still looks like it has a lot of prettier related changes that are not related to password changing

PR #2521, opened 2022-01-25
Fix get_variant_id

DrAnax:


Zamiell:
can we change it to not use an infinite loop

DrAnax:
Sure, I just translated Python to Typescript, but I can improve it here and there

PR #2522, opened 2022-01-25
Those cards are odd

### LINKED ISSUE #2299: [Variant] Odds & Evens
"Those 3 cards are odds."

DrAnax:
Closes #2299 

Props to Romain672 for the new variant

### DIFF
--- a/server/src/variants.go
+++ b/server/src/variants.go
@@ -257,6 +257,7 @@ func variantIsCardTouched(variantName string, clue Clue, card *Card) bool {
 		}
 
 		if variant.OddsAndEvens {
+			// Clue ranks in Odds And Evens can only be 1 or 2
 			if clue.Value == 1 {
 				return intInSlice(card.Rank, oddClues)
 			}
--- a/server/src/variants.go
+++ b/server/src/variants.go
@@ -14,6 +14,8 @@ var (
 	variants     map[string]*Variant
 	variantIDMap map[int]string
 	variantNames []string
+	oddClues     []int
+	evenClues    []int
 )
 
 // VariantJSON is very similar to Variant,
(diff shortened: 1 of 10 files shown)


server/src/variants.go
```diff
@@ -249,6 +256,13 @@ func variantIsCardTouched(variantName string, clue Clue, card *Card) bool {
 			return false
 		}
 
+		if variant.OddsAndEvens {
+			if clue.Value == 1 {
```
Zamiell, line 260: is there a constant for this

DrAnax: For `clue.Value == 1`? No

Zamiell: yes there is https://github.com/Hanabi-Live/hanabi-live/blob/main/server/src/constants.go#L32-L35

DrAnax: `clue.Value == 1` means that the player clued `1` to someone else. This is checked because in Odds And Evens a rank clue can be either `1` for odd or `2` for even ranked cards.

DrAnax: > yes there is https://github.com/Hanabi-Live/hanabi-live/blob/main/server/src/constants.go#L32-L35

This defines the clue.Type

Zamiell: oh i see

Zamiell:
this PR doesn't include documentation in variants.md

also, shouldn't the "== 1" be checking for any odd number instead of just number 1?

DrAnax:
> also, shouldn't the "== 1" be checking for any odd number instead of just number 1?

The rank clue in Odds and Evens can be either `1` or `2`. All other rank clues are off.

Zamiell:
ok, can you add that as a comment at the "== 1" place?

Zamiell:
thanks

PR #2528, opened 2022-01-27
Change password

DrAnax:
It's the same commit, I just undid by hand in main.tmpl what an extension I didn't choose to install and force into the codebase did.


Zamiell:
thanks

Zamiell:
can you do PRs now for
1) prettier on main.tmpl
2) refactoring http_login.go into smaller functions, right now the code is a mess

Issue #2533, opened 2022-01-28
Changing shared-missing to allow for players not to be taken into account


nitawa-hl:
Hi! re: [this](https://discord.com/channels/140016142600241152/733076625079402647/932718660022308924), I thought the the shared-missing URL could be changed slightly to include the number of players at the beginning, that way removing a player from the URL would still give the list of missing scores for the right amount of players. That is, instead of
/shared-missing/Alice/Bob/Cathy
write
/shared-missing/3/Alice/Bob/Cathy
and that way if Bob doesn't care about scorehunting, Alice and Cathy can just use
/shared-missing/3/Alice/Cathy

PS. I already have the PR ready, but I can't really test things on my PC at the moment, so any help with that part would be appreciated.

DrAnax:
I certainly can help you, publish on your fork so I can test it

DrAnax:
Code tested and produces correct results.

Zamiell:
user ids can be numbers so i dont see how this can work

nitawa-hl:
It doesn't have backwards compatibility, with the PR the address /shared-missing/Alice/Bob raises an error saying that you need to indicate the number of players first. I also changed the /missing command to add the number according to the number of players in the lobby

Zamiell:
i dont want to make urls harder to type to support your esoteric use-case

Zamiell:
> Did you see the discord conversation I linked to? 

no, website features should not be discussed on discord

> also, it literally only adds 2 chars to the url

i'm aware of that, "literally only 2 chars" makes the url literally harder to type

mmelwen:
your idea looks useful me to help the newer players score hunt, the way I do now it is with a blank user I replace my name in the url with the "dummy" user name

nitawa-hl:
I see this as an improvement to the use of a dummy user in serveral aspects:
- You don't need multiple dummy accounts to replace multiple players (personally, I only have one alt for teaching purposes)
- You won't have several players creating dummy accounts for this purpose
- It is more obvious to newer players how to remove someone from the missing: In a pre-game chat, if someone types /missing, it's easier (imo) to say "you can remove me from that" than "you can substitute me with X on the url" 
- Even better, it opens the door to adding some checkbox in the settings so that your account is removed from the missing command by default.

PR #2534, opened 2022-01-28
Change shared-missing to indicate number of players

nitawa-hl:
Solves Issue #2533

Thanks for the idea to Nitrate and for help testing it to Nitrate2 and @DrAnax :)


nitawa-hl:
Will try again

PR #2545, opened 2022-01-31
Odds and evens

### LINKED ISSUE #2544: make odds and evens less confusing
instead of a 1 clue and a 2 clue it should show to the play as O and E

this should apply to the clue arrows, and also the clue UI buttons

DrAnax:
Mentioned in #2299
Closes #2544

- Available clues, arrows and logs changed from 1/2 to Odd/Even
- Hypo parsing takes into account Odds And Evens variation
- Added special suit variations


Zamiell:
thank you very much

PR #2560, opened 2022-02-02
Support seeds with "legacy-x-" in front

DrAnax:


Zamiell:
what is the point of the len check

DrAnax:
Can't people put their own seed? This breaks if the seed is just `legacy-`

Zamiell:
ok this is a little confusing, but the user inputted seed is a suffix to the actual seed

meaning that it would be p2v5sLOLDICKS

DrAnax:
What if at some point we decide to rename again the seeds? I thought this would be a good safeguard

Zamiell:
well if we renamed the seeds we still wouldn't allow arbitrary prefixes

PR #2562, opened 2022-02-02
Disable /subscribe /unsubscribe /here /ping

DrAnax:


Zamiell:
probably dont have to remove the doc entries? since you are going to presumably put this back

DrAnax:
It's ok, I'll put the docs back when it's ready

Zamiell:
okie dokie

PR #2564, opened 2022-02-02
Special suits take precedence in clues before O&E

### LINKED ISSUE #2563: Special suits should override O&E clues
Pink should be touched by all clues, etc

DrAnax:
Closes #2563


Zamiell:
good thank you

PR #2566, opened 2022-02-03
Chat commands

DrAnax:
Already done:
- /here, /ping
- /subscribe, /unsubscribe
- Undocumented commands
- `/teachme`
- `/help` returns a PM with all the available chat commands and the github link
- One-line responses moved to external file
- Modal responses of commands are now PMs (except in case of system error)
- `/warning` command removed

Todo:
- replace client commands with slash messages
- create addToChat server response to avoid having commands back and forth from discord
- check for valid commands before outputting user's slash command (notify via PM)
- check for invalid commands currently (not in lobby/table, discord not ready etc) before outputting user's slash command (notify via PM)


Zamiell:
ok it looks like you are making a private commands thing.

in the past i haven't done that, i have just used "real" commands for that
for example, see what i did with the /tag command.

if you think that your solution is better, then i think that you should convert the other private commands (like /tag and so forth) to be using your custom system as well

Zamiell:
this is a tiny pr now :)

DrAnax:
Not yet

Zamiell:
close this now?

DrAnax:
Can't touch this

DrAnax:
look at the updated description

Zamiell:
> replace client commands with slash messages

is that a good idea? different commands have different data schemas

PR #2574, opened 2022-02-06
Compress hypo

### LINKED ISSUE #2499: Add endpoint to allow viewing JSON data
It could be either a GET to `/replay` with query var or a new POST to /replay

### LINKED ISSUE #2512: Add initial turn option to JSON replay
So that, if you share a json replay, it starts at the turn of interest when opened. (As far as I can tell the option is there for the !replay command, but not for the json replay.)

DrAnax:
Closes #2499 

Closes #2512

During hypo, `/copy` also tries to compress the JSON and create a URL.

If successful, it presents the URL to the user, as well as a button which copies the URL to the clipboard (replacing the JSON data).

image


DarthGandalf:
wouldn't it be simpler to compress the json string with zlib/zstd and pass the result through base64?

DrAnax:
> wouldn't it be simpler to compress the json string with zlib/zstd and pass the result through base64?

I tried many available solutions, nothing produced something that small.

Zamiell:
i think it should copy to clipboard by default, no need to force the user to click an additional button

DrAnax:
What if the user just wants the JSON data and not the URL?

DrAnax:
Maybe default into copying the URL to the clipboard and add the button to copy the raw JSON instead?

Zamiell:
also the function should return null on failure instead of empty string

Zamiell:
> Maybe default into copying the URL to the clipboard and add the button to copy the raw JSON instead?

sure

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ * Returns empty string if decompression fails.
+ *
+ * @param data The GameJSON object
+ */
+function gameJSONCompress(data: GameJSON): string {
+  let out = "";
+
+  // number of players
+  const numberOfPlayers = data.players.length;
+  if (numberOfPlayers < 2 || numberOfPlayers > 6) {
```
Zamiell, line 89: aren't there constants for this somewhere?

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ *
+ * @param src A compressed single character string representing a DeckCard
+ * @param range The validated min and max values of DeckCard.rank
+ */
+function stringToDeckCard(src: string, range: MinMax): DeckCard {
+  const r = range.max - range.min + 1;
+  const index = BASE62.indexOf(src);
+  const rank = (index % r) + 1;
+  const suitIndex = Math.floor((index - rank + 1) / r);
+  return <DeckCard>{ suitIndex, rank };
```
Zamiell, line 250: don't cast like this in typescript, use the `as` keyword instead

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ * of Action.type found in an array of Action
+ *
+ * @param actions The array of Action[] to search
+ */
+function getTypeMinMax(actions: Action[]): MinMax {
+  const range: MinMax = {
+    min: -1,
+    max: -1,
+  };
+  actions.forEach((action) => {
```
Zamiell, line 385: nit: the code base generally uses `for (const a of b)` over forEach

Zamiell:
why did you just base62 instead of the more-common base64

DarthGandalf:
at some point I'd like morphs in the result; but it looks like it should be possible to add them here without changing too much

DrAnax:
> why did you use base62 instead of the more-common base64

In the end it becomes base64 by adding hyphens. The encoding uses base62 to allow inserting hyphens without breaking the compress/decompression

DrAnax:
Done

DrAnax:
Now allows to create URL from partial replay (up to that point), without even using the hypo.

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
+import { isEqual } from "lodash";
+import { MAX_PLAYERS, MIN_PLAYERS } from ".";
+import { HYPO_PLAYER_NAMES } from "./hypoPlayerNames";
+import variantsJSON from "./json/variants.json";
+import { VariantJSON } from "./types/VariantJSON";
+
+type DeckCard = {
```
Zamiell, line 7: these 3 things should be interfaces, not types

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+  seed: string;
+};
+
+type MinMax = {
+  min: number;
+  max: number;
+};
+
+const BASE62 = "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+const variantsJSONArray = Array.from(variantsJSON) as VariantJSON[];
```
Zamiell, line 39: this is a bug, use getVariant instead

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ * Returns empty string if decompression fails.
+ *
+ * @param data The GameJSON object
+ */
+function gameJSONCompress(data: GameJSON): string {
+  let out = "";
+
+  // number of players
+  const numberOfPlayers = data.players.length;
+  if (numberOfPlayers < 2 || numberOfPlayers > 6) {
```
Zamiell, line 93: you aren't using the constants you made

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+    return "";
+  }
+  out += `${data.players.length}`;
+
+  // deck
```
Zamiell, line 98: comments should have capital letters

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ * Decompresses a string into a GameJSON object.
+ * Returns empty string if decompression fails.
+ *
+ * @param data The compressed string
+ */
+function gameJSONDecompress(data: string): string {
+  // remove all hyphens from URL
+  const normal = data.replace(/-/g, "");
+
+  // The compressed string is composed of 3 substrings, separated by comma.
```
Zamiell, line 134: comments should not have periods

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+  const [deckString, actionsString, variantString] = [...normal.split(",", 3)];
+  const numPlayers = parseInt(deckString.charAt(0), 10);
+
+  const players = getPlayers(numPlayers);
+  if (players.length === 0) {
+    return "";
```
Zamiell, line 140: return null on failure instead of an empty string

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+      range.min === -1 ? action.type : Math.min(range.min, action.type);
+    range.max =
+      range.max === -1 ? action.type : Math.max(range.max, action.type);
+  }
+
+  return range;
+}
+
+/**
+ * Ensures that Minmax.min is non negative and smaller or equal to Minmax.max
```
Zamiell, line 396: jsdoc must have full sentences

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,402 @@
...
+ * Decompresses a string into an array of DeckCard.
+ * The first two characters of the string must be numbers containing the min and max values of DeckCard.rank.
+ * The rest of the string must be composed of series of strings one character long, each representing a DeckCard.
+ *
+ * @param src The compressed string
+ */
+function decompressDeck(src: string): [DeckCard[], boolean] {
+  const deck: DeckCard[] = [];
+  const range: MinMax = {
+    min: parseInt(src.charAt(0), 10),
```
Zamiell, line 204: use parseintsafe instead of parseint, also check for isNaN, also use SLR

DrAnax: Why use parseIntSafe since I'm only parsing a single character string?

Zamiell: because parseint sucks and you should never use it

DrAnax:
After this one is accepted, #2575 will probably have merge conflicts

DrAnax:
This should pass after #2579 is accepted and tests are rerun.

packages/data/src/hypoCompress.ts
DrAnax: It's a female comment

Zamiell:
looks like it doesn't pass

DrAnax:
Yes due to last rename. Fixed

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,51 @@ export default function createJSONFromReplay(room: string) {
...
-    room,
-  );
+  const URLData = shrink(json);
+  if (URLData === "" || URLData === null) {
+    chat.addSelf(
+      '<span class="red">Error</span>: Your JSON data cannot be compressed.',
+      room,
+    );
+  } else {
+    const URL = `https://hanab.live/shared-replay-json/${URLData}`;
```
Zamiell, line 79: shouldn't this be
```
const URL = `/shared-replay-json/${URLData}`;
```
instead?

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,51 @@ export default function createJSONFromReplay(room: string) {
...
+    navigator.clipboard.writeText(URL).then(
```
Zamiell, line 80: why did you write .then here?

DrAnax: because navigator.clipboard.writeText returns a promise.

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,51 @@ export default function createJSONFromReplay(room: string) {
...
+      () => {},
+      () => {},
+    );
+    chat.addSelf(
+      '<span class="green">Info</span>: Your hypo URL is copied on your clipboard.',
+      room,
+    );
+    const here = `<button href="#" onclick="navigator.clipboard.writeText('${json.replace(
```
Zamiell, line 88: 1) violates SLR
2) using .then for no reason
3) it should be </strong> instead of < /strong>
4) it should be </button> instead of </button >

packages/client/src/lobby/createReplayJSON.ts
DrAnax: no because the url in the clipboard would be `/shared-replay-json/...`

packages/data/src/hypoCompress.ts
Zamiell: why was this marked as resolved when they are still types

packages/data/src/hypoCompress.ts
Zamiell: why was this marked as resolved when it wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/data/src/hypoCompress.ts
Zamiell: marked as resolved but wasn't fixed

packages/client/src/lobby/createReplayJSON.ts
Zamiell: just because it returns a promise doesn't mean you need to use an empty then

packages/client/src/lobby/createReplayJSON.ts
Zamiell: ok, then don't hard-code the protocol/domain and get it from the proper constant(s)

DrAnax:
All the resolved issues where fixed, I don't know why this happened; redoing...

packages/client/src/lobby/createReplayJSON.ts
Zamiell: this was marked as resolved but it isn't resolved

packages/data/src/constants.ts
```diff
@@ -10,3 +10,5 @@ export const DEFAULT_CLUE_RANKS: readonly number[] = [1, 2, 3, 4, 5];
 
 export const MIN_PLAYERS = 2;
 export const MAX_PLAYERS = 6;
+
+export const MAX_CARDS_IN_A_DECK = 60;
```
Zamiell, line 13: should probably be a composition of MAX_CARDS_IN_A_SUIT and MAX_SUITS_IN_A_VARIANT

packages/data/src/hypoCompress.ts
```diff
@@ -1,25 +1,34 @@
...
-
-type DeckCard = {
+import {
+  getVariant,
+  getVariantByID,
+  HYPO_PLAYER_NAMES,
+  MAX_PLAYERS,
+  MIN_PLAYERS,
+  parseIntSafe,
+} from ".";
```
Zamiell, line 9: bug

packages/data/src/hypoCompress.ts
```diff
@@ -69,97 +84,41 @@ export function shrink(JSONString: string): string | null {
...
-  const numPlayers = parseInt(deckString.charAt(0), 10);
+  // The compressed string is composed of 3 substrings, separated by comma
+  const [deckString, actionsString, variantID] = [...normal.split(",", 3)];
+  const numberPlayersString = deckString.charAt(0);
+  const numPlayers = parseIntSafe(numberPlayersString);
 
-  const players = getPlayers(numPlayers);
-  if (players.length === 0) {
-    return "";
+  const playersArray = getPlayers(numPlayers);
```
Zamiell, line 102: shouldn't it be called players so that it matches the getPlayers function?

packages/data/src/hypoCompress.ts
```diff
@@ -69,97 +84,41 @@ export function shrink(JSONString: string): string | null {
...
+  if (playersArray.length === 0) {
+    return null;
   }
 
-  const [deck, okDeck] = decompressDeck(deckString.substring(1));
-  if (!okDeck) {
-    return "";
+  const deck = decompressDeck(deckString.substring(1));
```
Zamiell, line 107: SLR

packages/data/src/hypoCompress.ts
DrAnax: what's the bug?

Zamiell: `from "."` doesn't make any sense

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,57 @@ export default function createJSONFromReplay(room: string) {
...
-  const log = states[states.length - 1].log;
-  if (replay.segment < log.length) {
-    const slice = log.slice(replay.segment + 1);
-    const actions = getGameActionsFromLog(slice);
-    game.actions.push(...actions);
+  if (replay.hypothetical !== null) {
+    const states = replay.hypothetical.states;
+    const log = states[states.length - 1].log;
+    if (replay.segment < log.length) {
+      const slice = log.slice(replay.segment + 1);
```
Zamiell, line 56: slice is a bad name, the variable name should describe what the data actually contains

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,57 @@ export default function createJSONFromReplay(room: string) {
...
 
   // Add the hypothesis from log, after current segment
-  const states = replay.hypothetical!.states;
-  const log = states[states.length - 1].log;
-  if (replay.segment < log.length) {
-    const slice = log.slice(replay.segment + 1);
-    const actions = getGameActionsFromLog(slice);
-    game.actions.push(...actions);
+  if (replay.hypothetical !== null) {
+    const states = replay.hypothetical.states;
```
Zamiell, line 53: i think destructuring would be better here

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -50,23 +49,57 @@ export default function createJSONFromReplay(room: string) {
...
+  const URL = `${SITE_URL}/shared-replay-json/${URLData}`;
+  navigator.clipboard
+    .writeText(URL)
+    .then(() => {
+      chat.addSelf(
+        '<span class="green">Info</span>: Your hypo URL is copied on your clipboard.',
+        room,
+      );
+      const urlFix = json.replace(/"/g, "\\'");
+      const here = `<button href="#" onclick="navigator.clipboard.writeText(${urlFix}.replace(/\\'/g, String.fromCharCode(34))).then(()=>{},()=>{});return false;"><strong>here</strong></button>`;
```
Zamiell, line 90: get rid of .then

Zamiell: and use SLR, this line is disgusting

packages/client/src/lobby/createReplayJSON.ts
DrAnax: not sure what to do

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+ * Returns null if the compression fails.
+ *
+ * The resulting string is composed of three substrings separated by comma.
+ * The first substring represents the number of players and the deck.
+ * The second substring represents the actions.
+ * The third substring is the ID of the variant.
+ *
+ * Finally hyphens are added to the string to make URL text wrap when posting.
+ *
+ * @param JSONString A string representing a hypo JSON
```
Zamiell, line 50: jsodc has to have complete sentences

packages/client/src/lobby/createReplayJSON.ts
Zamiell: `const { states } = replay.hypothetical;`

Zamiell: the point is that you avoid typing the word states twice, which wastes time + creates more code

packages/client/src/lobby/createReplayJSON.ts
Zamiell: its marked as resolved but i dont see it changed

packages/data/src/gameData.ts
```diff
@@ -34,6 +42,21 @@ export function getVariant(variantName: string): Variant {
   return variant;
 }
 
+export function getVariantByID(id: number | string): Variant {
```
Zamiell, line 45: change function signature to only take a number, and then use an ID index that we create on init

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  return compressed;
+}
+
+/**
+ * Decompresses a string into a GameJSON object.
+ * Returns null if decompression fails.
+ *
+ * @param data The compressed string
```
Zamiell, line 86: jsdoc must be complete sentances

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+ */
+export function expand(data: string): string | null {
+  // Remove all hyphens from URL
+  const normal = data.replace(/-/g, "");
+
+  // The compressed string is composed of 3 substrings, separated by comma
```
Zamiell, line 92: should be "commas" instead of "comma"

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+  const suitIndex = Math.floor((index - rank + 1) / r);
+  return { suitIndex, rank };
+}
+
+/**
+ * Compresses an array of Action into a string.
+ * The first two characters of the string are numbers containing the min and max values of Action.type.
+ * The rest of the string contains series of strings two characters long, each representing an Action.
+ *
+ * @param actions The array of Action
```
Zamiell, line 258: jsdoc must have complete sentances

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+function compressActions(actions: Action[]): string {
+  // Find min/max values of Action.type
+  const typeRange = getTypeMinMax(actions);
+  if (isMinMaxInvalid(typeRange)) {
+    // Empty actions still require min/max values
+    return "00";
+  }
+
+  let out = `${typeRange.min}${typeRange.max}`;
+  actions.forEach((action) => {
```
Zamiell, line 269: use a `for each` loop instead of forEach to match existing style

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+
+  return out;
+}
+
+/**
+ * Decompresses a string into an array of action.
+ * The first two characters of the string must be numbers containing the min and max values of Action.type.
+ * The rest of the string must be composed of series of strings two characters long, each representing an Action.
+ *
+ * @param src The compressed string
```
Zamiell, line 281: complete sentances

packages/data/src/hypoCompress.ts
```diff
@@ -0,0 +1,406 @@
...
+  return actions;
+}
+
+/**
+ * Converts an Action into a string.
+ * The returned string is two characters long,
+ * the first one representing Action.type and Action.value
+ * and the second representing the Action.target.
+ *
+ * @param action The Action object
```
Zamiell, line 312: same here

packages/client/src/lobby/createReplayJSON.ts
DrAnax: I haven't sent my commit yet. I'm fixing and marking as I go

Zamiell: do you do it in order from top to bottom? if yes, then maybe its better to wait until the commit is live, so that i don't get confused

packages/data/src/hypoCompress.ts
DrAnax: ?

Zamiell: its missing a period at the end, also use "hypothetical" instead of "hypo" to avoid confusion.

Zamiell: (see https://jsdoc.app/tags-example.html)

packages/data/src/hypoCompress.ts
DrAnax: Actually it should be commata (greek word, plural) but I'll settle for commas.

Zamiell: =p

packages/data/src/hypoCompress.ts
```diff
@@ -89,8 +89,10 @@ export function expand(data: string): string | null {
   // Remove all hyphens from URL
   const normal = data.replace(/-/g, "");
 
-  // The compressed string is composed of 3 substrings, separated by comma
-  const [playersAndDeck, actionsString, variantID] = [...normal.split(",", 3)];
+  // The compressed string is composed of 3 substrings, separated by commas.
```
Zamiell, line 92: don't use punctuation in nonjsdoc comments

packages/client/src/lobby/hypoCompress.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  try {
+    const variant = getVariant(data.options.variant);
+    out += `${variant.id}`;
+  } catch {
+    return null;
+  }
+
+  // Add hyphens every 20 characters for URL posting (hyphens make the text wrap)
+  out = out.match(/.{1,20}/g)!.join("-");
```
Zamiell, line 179: shouldn't we check for the error instead of using a null assertion operator?

DrAnax: at that point, out is certainly non empty or the code would have returned null previously

Zamiell: if you change any code, then out won't be guaranteed anymore to be 20 characters long, so it is fragile

DrAnax: it doesn't have to. If out is less than 20 characters long, match will return an array with a single element (the entire out) and join will do nothing (effectively, with less than 20 characters no hyphen is added)

Zamiell: then why is the null assertion operator necessary to have in there at all if the code will never be null

DrAnax: because the compiler sees `match` and believes it might find nothing, which can never be the case here

Zamiell: ok

Zamiell:
nice work! this is a very complicated feature

PR #2575, opened 2022-02-07
Replace modals with PM, remove /warning

DrAnax:
- Replace modals when the user types a command with PM.- 
- Remove `/warning` (probably made for debug purposes)- 
- Add chat.addSelf message types.- 
- Replace "that command" with specific commands.


packages/client/src/chat.ts
```diff
@@ -589,16 +589,30 @@ export function add(data: ChatMessage, fast: boolean): void {
 }
 
 // addSelf is used when the client needs to send a chat message to itself
-export function addSelf(msg: string, room: string): void {
+export function addSelf(
+  msg: string,
+  room: string,
+  type: "normal" | "info" | "error" = "normal",
```
Zamiell, line 595: use an enum

packages/client/src/chat.ts
```diff
@@ -589,16 +589,30 @@ export function add(data: ChatMessage, fast: boolean): void {
...
+): void {
+  let message = msg;
```
Zamiell, line 597: make a getMessage function

packages/client/src/chat.ts
```diff
@@ -12,6 +12,12 @@ import { parseIntSafe } from "./misc";
 import * as modals from "./modals";
 import ChatMessage from "./types/ChatMessage";
 
+export enum SelfChatMessageType {
+  normal,
```
Zamiell, line 16: in typescript, the values of an enum should be capitalized, see: https://www.typescriptlang.org/docs/handbook/enums.html

packages/client/src/chat.ts
```diff
@@ -589,16 +595,21 @@ export function add(data: ChatMessage, fast: boolean): void {
 }
 
 // addSelf is used when the client needs to send a chat message to itself
-export function addSelf(msg: string, room: string): void {
+export function addSelf(
+  msg: string,
+  room: string,
+  type: SelfChatMessageType = SelfChatMessageType.normal,
```
Zamiell, line 601: the type annotation is superfluous

packages/client/src/chat.ts
```diff
@@ -589,16 +595,21 @@ export function add(data: ChatMessage, fast: boolean): void {
...
+): void {
+  const message = getChatMessage(msg, type);
   add(
     {
-      msg,
-      who: "",
+      msg: message,
+      who: "Hanab Live",
```
Zamiell, line 607: my bet is that this string is a constant that lives somewhere else, and if it is, then it shouldn't be hardcoded here

packages/client/src/chat.ts
```diff
@@ -699,3 +710,17 @@ export function updatePeopleTyping(): void {
   chat1.html(msg);
   chat2.html(msg);
 }
+
+function getChatMessage(msg: string, type: SelfChatMessageType): string {
+  let message = msg;
+  switch (type) {
+    case SelfChatMessageType.info:
```
Zamiell, line 717: please use { } brackets around the cases of all switch statements to match the existing code style

packages/client/src/chat.ts
```diff
@@ -699,3 +710,17 @@ export function updatePeopleTyping(): void {
...
+      message = `<span class="green">${message}</span>`;
+      break;
+    case SelfChatMessageType.error:
+      message = `<span class="red">${message}</span>`;
+      break;
+    default:
```
Zamiell, line 723: use ensure all cases

packages/client/src/lobby/createReplayJSON.ts
```diff
@@ -63,9 +64,10 @@ export default function createJSONFromReplay(room: string) {
     () => {},
```
Zamiell, line 64: get rid of the then crap

packages/client/src/chat.ts
DrAnax: it's unreachable (`server_old/src/...` and `constants.go`), putting it into /data

packages/client/src/lobby/createReplayJSON.ts
DrAnax: need to accept either this or #2574 first so I known where to merge from

Zamiell: up to you, which one do you want to focus on first

packages/client/src/chat.ts
DrAnax: in default there's no message change but code could reach that point; should I add a simple break?

Zamiell: ill add it

packages/client/src/lobby/createReplayJSON.ts
DrAnax: accept the other one first please.

packages/client/src/chat.ts
Zamiell: the point of the function is to return values, not modify variables, get rid of the let

Zamiell:
thanks

PR #2576, opened 2022-02-07
One line replies

DrAnax:
- Puts all simple responses into an external file.
- Lists all available commands when using `/help` before the link message.


Zamiell:
good, thanks

PR #2579, opened 2022-02-08
Removes duplicate ensureAllCases and parseIntSafe

DrAnax:


Zamiell:
i googled util vs utils and it seems like consensus is with the latter

packages/data/src/console.ts
```diff
@@ -0,0 +1,4 @@
+export function error(message: string): never {
```
Zamiell, line 1: can we put this back in utils? it probably doesn't need to be in a separate file

DrAnax: It's only used when creating variants; ts-prune complains it's exported and never used.

Zamiell: you mean eslint?

Zamiell: oh i see

Zamiell: just stick it in createVariants.json then please

DrAnax: it's also used in newID.ts

Zamiell: can you put it in utils.ts then and add a ts-prune-ignore comment or whatever

Zamiell: also i think we should put all of the files related to creating variants in a subdirectory called `createVariants` to stay more organized

DrAnax: I put it in createVariants.ts, export it from there and importing it in newID.ts

Zamiell: won't that create a dep cycle?
maybe we should create utils.ts in the new createVariants subdirectory

DrAnax: All checks pass. Want me to create a new utils.ts?

Zamiell: clarification: all checks pass because in this repo there are circular deps, so i disabled that lint rule, but in general we should avoid circular deps in any new code we write

Zamiell: since circular deps are spaghetti

Zamiell:
good PR!!!!
thank you!

PR #2590, opened 2022-02-09
Fix highlight in Odds and Evens so it colors all odd/even numbers.

nitawa-hl:
Fixes this:
image
so it highlights all odd/even numbers.

Thanks @DrAnax for the help :)


Zamiell:
thx

Issue #1039, opened 2020-05-13
positive color clues should light up pip in gold or something


Zamiell:

DrAnax:
This should no longer be an issue

Zamiell:
why wouldn't it

DrAnax:
Why would you need gold pips here?

![image](https://user-images.githubusercontent.com/2622043/153019697-c68a009b-527b-4183-b0dd-2b8b8bc464cf.png)

Zamiell:
your screenshot isn't my concern

1) watch what happens to a pink card when it goes from having no clues on it to having one rank 2 clue on it
2) watch what happens to a rainbow-ones card when it goes from having no clues on it to having one blue clue on it

these two situations should have symmetry, and they currently do not

Romain672:
Here is another option, make colors bigger:
![big](https://user-images.githubusercontent.com/12693191/153283263-2ccc8788-ce6c-49f2-9a4c-837756a15341.jpg)

Zamiell:
making them bigger is kind of ugly, so i suspect that making them have a gold border would be better
(although I have not seen a screenshot yet with them having a gold border either)

Romain672:
Or a line which start from the center into the color clued.
Maybe an arrow.

DrAnax:
Better?

![image](https://user-images.githubusercontent.com/2622043/153507199-61532657-eaa4-428d-ac03-27144517fb1d.png)

DarthGandalf:
It looks very heavy

DarthGandalf:
at very least, no shadow behind the circle

the circle itself is huge

making the color more subtle might help, dunno

try the gold border instead of circle?

DarthGandalf:
Can you share the code what you did so far to plumb the parameters? I can also try to poke the graphics a bit

DrAnax:
omw

DrAnax:
drawPips.ts 3370

DrAnax:
I think the Rainbow pip should never be highlighted because there's no "Rainbow" Color clue. Highlight (after this PR) means "positive information" and PO "Rainbow" can never be given.

DrAnax:
Their respective pips should; clues can get them.

DrAnax:
There's no "mahogany" pip, only red and purple

nitawa-hl:
Ok, then I think I have it, give me a sec and I push again to my fork so you can test it if you want :)

[...]

PR #2593, opened 2022-02-10
Positive color clues

DrAnax:
Closes: #1039

### DIFF
--- a/packages/client/src/game/types/Pips.ts
+++ b/packages/client/src/game/types/Pips.ts
@@ -3,7 +3,9 @@ import RankPip from "../ui/controls/RankPip";
 
 export default interface Pips {
   suitPips: Konva.Group;
+  suitPipsPositive: Konva.Group;
   suitPipsMap: Map<number, Konva.Shape>;
+  suitPipsPositiveMap: Map<number, Konva.Shape>;
   suitPipsXMap: Map<number, Konva.Shape>;
   rankPips: Konva.Group;
   rankPipsMap: Map<number, RankPip>;
--- a/packages/client/src/game/ui/HanabiCard.ts
+++ b/packages/client/src/game/ui/HanabiCard.ts
@@ -68,8 +68,10 @@ export default class HanabiCard
   private finesseBorder: Konva.Group;
 
   private suitPips: Konva.Group;
+  private suitPipsPositive: Konva.Group;
   private rankPips: Konva.Group;
   private suitPipsMap: Map<number, Konva.Shape>;
+  private suitPipsPositiveMap: Map<number, Konva.Shape>;
   private suitPipsXMap: Map<number, Konva.Shape>;
   private rankPipsMap: Map<number, RankPip>;
   private rankPipsXMap: Map<number, Konva.Shape>;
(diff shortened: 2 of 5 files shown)


DarthGandalf:
3 clues were given; the cards turned into known 1s after the second color clue, but the 3rd clue didn't update the pip

image

DrAnax:
I tried altering the border, but it didn't always look right (or wasn't visible enough). Here's the dinal render in various resolutions:

image

image

image

DarthGandalf:
a more subtle one:

изображение
изображение

DrAnax:
> a more subtle one:

Try it also in smaller resolutions

DarthGandalf:
изображение
изображение
изображение

DarthGandalf:
here's with border instead:

изображение
изображение

but it's barely visible on white background

DrAnax:
> here's with border instead:
> but it's barely visible on white background

And if you try it in smaller resolution it's next to invisible.

DrAnax:
There, either solution with a circle works, it's a matter of preference now.

DarthGandalf:
here's code for the subtle highlight I posted above:

```js
  if (highLight === true) {
    ctx.save();
    ctx.beginPath();
    ctx.arc(3, 5, 130, 0, 360);
    ctx.closePath();
    const gradient = ctx.createRadialGradient(0,0,3, 0,0,130);
    gradient.addColorStop(0, 'rgba(238, 188, 29, 0.7)');
    gradient.addColorStop(0.4, 'rgba(238, 188, 29, 0.5)');
    gradient.addColorStop(0.7, 'rgba(238, 188, 29, 0.2)');
    gradient.addColorStop(1, 'rgba(238, 188, 29, 0.0)');
    ctx.fillStyle = gradient;
    ctx.shadowColor = "rgba(0, 0, 0, 0)";
    ctx.fill();
    ctx.restore();
  }
```

nitawa-hl:
I tried something! I think it looks quite good, except that the pips/shadows are doing weird things, and that the blue pip overlaps with the 1, but just to give you an idea before I spend any more time on it...

image
image
image

nitawa-hl:
The issue with the yellow pip comes from the shapeFunction, see here: https://jsbin.com/katofenapi/edit?js,output

Zamiell:
ok so which is one is the best one out of all the ones that people posted

DrAnax:
> ok so which is one is the best one out of all the ones that people posted

We have rested our case, it's up to you to decide.

Zamiell:
which one do you like

DarthGandalf:
Let's do this:

who wants the [full circle behind](https://user-images.githubusercontent.com/325092/153515324-ba58cb55-5b6e-43e1-994d-50e0c8fc08b0.png), react with Rocket
who wants [gradient circle behind](https://user-images.githubusercontent.com/325092/153522987-193de8f5-ca8c-4ee6-82de-3efd7dd5dcbd.png), react with Heart
who wants the [border](https://user-images.githubusercontent.com/83447589/153680192-d45f01a9-5301-4709-8fb9-af874649844f.png), react with Eyes

DrAnax:
I'm not a good candidate to ask because of my poor eyesight (a lot of presbyopia). Plus I would be biased towards what I did.

Romain672:
Since it's close, I would not said no to a screenshot with thrice the same hand. And then for differents variants (synesthesia & X-Ones?).

DarthGandalf:
@Zamiell looks like the border got 1 more vote than circle, so let's merge #2607

PR #2612, opened 2022-02-13
Make cluing UI more obvious

### LINKED ISSUE #2075: Confusing UI of giving clue
Currently there are 2 independent lists of players. When I want to give a clue to someone, I need to look at their name, find that name in the other list (above numbers/colors selector for the clue), and select it there.

Why not make the list to be the same? E.g. move these buttons from above the numbers/colors to be near the players' hands. I can draw a mock UI if you want.

At first I couldn't figure out how to give the clue: I haven't noticed the other list at all, and I tried to click the player's name near their hand, which only shown a log of what that player did so far. When I introduced my group to this website, (they already knew hanabi both from real life and from BGA), they also were confused how to give the clue.

DarthGandalf:
Fix #2075

### DIFF
--- a/packages/client/src/game/ui/controls/PlayerButton.ts
+++ b/packages/client/src/game/ui/controls/PlayerButton.ts
@@ -3,9 +3,37 @@ import Button from "./Button";
 
 export default class PlayerButton extends Button {
   targetIndex: number;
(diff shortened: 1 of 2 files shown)


Zamiell:
thanks

PR #2614, opened 2022-02-16
Update sample .ts file

calvinballing:
Changes sample .ts file from `public\js\src\main.ts` to `Client\src\main.ts` since there is no longer a file at the first location


Zamiell:
thanks

PR #2621, opened 2022-03-04
Fix Hanabi type in RULES.MD

zionts:


Zamiell:
is it already spelled correctly

Issue #2625, opened 2022-04-24
Inappropriate license change


DarthGandalf:
bddce95c97a427ff079ed7064362a2edf643dd6c changed license of the project from GPL to AGPL without getting permission of all the copyright holders.

Section 13 of GPLv3 allows using the GPL-licensed code together with AGPL code, but doesn't allow changing the license completely.

It is possible to change license, but that's not how you do it.

Please revert.

Zamiell:
can you submit a PR

PR #2626, opened 2022-04-24
Set license back from AGPL to GPL

DarthGandalf:
Partially revert bddce95c97a427ff079ed7064362a2edf643dd6c

See #2625


timotree3:
Don't you need to get permission from all the people who have contributed since the license was changed to AGPL?

Zamiell:
lul

DarthGandalf:
Yep, you can do that. I'd prefer James to do that though, since he's the one who created this mess, sorry.

PR #2637, opened 2022-06-11
Positive color clues (fixed)

DarthGandalf:
Brings #2593 back, with X


DarthGandalf:
Выделение_430Выделение_429

Zamiell:
Thank you sir

PR #2647, opened 2022-07-02
Update getVariantDescriptions.ts

gisymon:
Removing 3-suit variants that are duplicates with Ambiguous combos.

### DIFF
--- a/packages/data/src/getVariantDescriptions.ts
+++ b/packages/data/src/getVariantDescriptions.ts
@@ -20,7 +20,7 @@ const SUITS_THAT_CAUSE_DUPLICATED_VARIANTS_WITH_AMBIGUOUS = new Set<string>([
   "Dark Prism", // This is the same as Dark Rainbow
 ]);
 
-const SUITS_THAT_REQUIRE_TWO_CLUEABLE_SUITS = new Set<String>([
+const SUITS_THAT_REQUIRE_TWO_CLUEABLE_SUITS = new Set<string>([
   "Rainbow",
   "Prism",
 ]);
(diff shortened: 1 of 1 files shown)


Zamiell:
@timotree3 is this right?

vEnhance:
https://github.com/gisymon/hanabi-live/pull/1 to pass lint

Zamiell:
@gisymon can you accept the pr?

timotree3:
@vEnhance's PR https://github.com/gisymon/hanabi-live/pull/1 was onto gisymon:main not gisymon:patch-1 so it didn't affect this PR

packages/data/src/getVariantDescriptions.ts
```diff
@@ -138,6 +143,17 @@ export function getVariantsForEachSpecialSuitCombination(
           continue;
         }
 
+        // Prism and Rainbow require 2 clueable suits, else they are just ambiguous red
+        if (
+          numSuits === 3 &&
+          ((SUITS_THAT_REQUIRE_TWO_CLUEABLE_SUITS.has(suit.name) &&
+            (suit2.noClueColors === true || suit2.allClueColors === true)) ||
+            ((suit.noClueColors === true || suit.allClueColors === true) &&
```
timotree3, line 151: From what I can tell, these should be booleans, so I don't get why the [`strict-boolean-expressions`](https://typescript-eslint.io/rules/strict-boolean-expressions/) lint was going off.

https://github.com/Hanabi-Live/hanabi-live/blob/ae74f7fbae505f0309f680e37980f043b0d79975/packages/data/src/types/Suit.ts#L16-L17

timotree3 (approved):
Code change seems to make sense. Thanks so much @gisymon and @vEnhance! This does mean that replays of the old variants will probably have to be deleted. It's probably worth giving people some time to export any that they care about before merging this

Zamiell:
@vEnhance can you just resubmit this as a new PR for gisymon? (since your PR to his PR didn't work)

gisymon:
Let me know if I need to do anything to get this through. Thanks!

Zamiell:
we will merge #2692 after we figure out the problem with the other two PRs

PR #2665, opened 2022-08-21
Update VARIANTS.md

vEnhance:


Zamiell:
thanks, the linter would have caught this on the other repo haha

PR #2678, opened 2022-10-06
Resolves #2643

### LINKED ISSUE #2643: Add any sign in the log when Hypothetical start
Like just a ">" would be enough.

vEnhance:
Adds "[Hypo]" in front of hypo actions in the replay, closes #2643


Zamiell:
looks good except it doesn't pass ci

Zamiell:
thanks! ill rebuild the js now on the site

PR #2680, opened 2022-10-06
Note bug fixes, turn count added

newduke:
* Fixed bug created by HTML parsing notes: "<foo" would not show anything.
* Fixed bug with meaningful notes getting surrounded by brackets: "[k3] some text" -> "[[k3] some text] [cm]"
* Added automated turn count to notes with ctrl + alt + right-click

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,10 @@ motd.txt*
 /logs
 /ongoing_tables
 
+# Development files
+*.code-workspace
+package-lock.json
+
 # Private keys
 GitHub_private_key.pem
 
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ motd.txt*
 
 # Development files
 *.code-workspace
-package-lock.json
 
 # Private keys
 GitHub_private_key.pem
--- a/packages/data/src/abbreviations.ts
+++ b/packages/data/src/abbreviations.ts
@@ -21,7 +21,7 @@ export const CHOP_MOVED_NOTES = [
 export const FINESSED_NOTES = ["f", "hf", "pf", "gd", "utf"];
 export const NEEDS_FIX_NOTES = ["fix", "fixme", "needs fix"];
 export const BLANK_NOTES = ["blank"];
-export const CLUED_NOTES = ["clued", "y"];
+export const CLUED_NOTES = ["clued", "yy"];
 export const UNCLUED_NOTES = ["unclued", "x"];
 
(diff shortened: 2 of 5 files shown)


packages/data/src/abbreviations.ts
```diff
@@ -21,7 +21,7 @@ export const CHOP_MOVED_NOTES = [
 export const FINESSED_NOTES = ["f", "hf", "pf", "gd", "utf"];
 export const NEEDS_FIX_NOTES = ["fix", "fixme", "needs fix"];
 export const BLANK_NOTES = ["blank"];
-export const CLUED_NOTES = ["clued"];
+export const CLUED_NOTES = ["clued", "y"];
```
Zamiell, line 24: y is short for yellow so this seems like a bug?

Zamiell:
> Added automated turn count to notes with ctrl + alt + right-click

that seems like a really good feature

packages/data/src/abbreviations.ts
newduke: Convenience? :)

.gitignore
```diff
@@ -30,6 +30,10 @@ motd.txt*
 /logs
 /ongoing_tables
 
+# Development files
+*.code-workspace
+package-lock.json
```
Zamiell, line 35: add this is a bug because we use yarn, right?

newduke: I run the build as instructed for linux.  had to `npm install` first to set the lock file.

Zamiell: thats old info, you shouldn't ever have to invoke npm

Zamiell: on this page it doesn't say anything about doing `npm install` https://github.com/Hanabi-Live/hanabi-live/blob/main/docs/INSTALL.md

newduke: That being said, it did not work without first running `npm install` because it complained about a missing lockfile

Zamiell: its because of this bug https://github.com/Hanabi-Live/hanabi-live/commit/37631cb77966973327eb904d113dce2df4fe1083

Zamiell:
should be live now, thanks

Zamiell:
oh, you should submit a pr that documents the new changes

Zamiell:
on features.md or whatever

newduke:
Wow that's fast. Does it take the bundles from CI then?

Zamiell:
what?

newduke:
Where does the live site get the javascript?  From gh ci?

Zamiell:
no i type git pull

PR #2691, opened 2022-10-20
Zen

### LINKED ISSUE #2645: Zen/tournament mode
Would it be possible to have a zen/tournament mode, wherein you can disable chat notifications and seeing the fun spectator eyes? Or at least disable spec commentary. Basically show only game state information. It would be nice to have that option for tournaments or when I really want to concentrate.

vEnhance:
Closes #2645. Activated by right-clicking the chat icon or by using /zen command in game.


Zamiell:
very cool thanks

PR #2692, opened 2022-10-21
Closes #2647

vEnhance:
- Fixes lint errors
- Also, actually delete the extraneous variants from variants.json and variants.txt
- Also, in 3a145984 I altered the variant JSON generation script so that any new variants are numbered starting after the highest number variant --- since I worried that re-using ID's of deleted variants later on could cause problems. If this is undesired and it's more important to have consecutive ID numbers for some reason, that commit should be reverted.


Zamiell:
i think it was important to consecutive ids but i dont remember why right now

also see discord dms

vEnhance:
> i think it was important to consecutive ids but i dont remember why right now

OK, I reverted that commit then. But we'll have some gaps once this is merged anyways, because we're deleting some existing variants.

> also see discord dms

:heavy_check_mark:

Zamiell:
@vEnhance one more thing, can you give me a list of the ids that were deleted?
so I can delete them from the database (at the time that i reboot the server with this new PR)

vEnhance:
> @vEnhance one more thing, can you give me a list of the numerical ids that were deleted? so I can delete them from the database (at the time that i reboot the server with this new PR)
> 
> e.g. i will type
> 
> ```
> DELETE FROM games WHERE id='1' OR id='2' OR id='3';
> ```

Look at the red lines in:

https://github.com/Hanabi-Live/hanabi-live/pull/2692/files#diff-46d40b68fad7ae57d7fef56de85728aab0a7b2ac75c55e0b4f68c567a34a0796

PR #2694, opened 2022-10-24
VARIANTS.md: fix suit order

chennisden:
The 5 suit colours are listed in the old order (before they were made into rainbow order rygbp). Probably doesn't matter, but thought I'd point it out and fix it when I saw it.


Zamiell:
Thanks

PR #2699, opened 2022-10-28
Adding /suggest to gameCommands

Krycke:
Adding /suggest to gameCommands
to be dispached to server
instead of parsed out in the chat.ts


Zamiell:
thanks

PR #2700, opened 2022-10-28
Zen mode but only on the client

vEnhance:
If this causes a deadlock then, uh, I have no idea, since it only touches the client.


Zamiell:
can you keep the npx?
there's no yarn equivalent of npx so its idiomatic in typescript projects to use it even if you are not using npm

packages/client/src/globals.ts
```diff
@@ -58,6 +58,8 @@ export class Globals {
   ui: HanabiUI | null = null;
   // Used to keep track of how many in-game chat messages are currently unread.
   chatUnread = 0;
+  // Zen mode
```
Zamiell, line 61: get rid of the comment since it doesn't explain anything

Zamiell: or alternatively make it explain what zen mode is

vEnhance:
> can you keep the npx? there's no yarn equivalent of npx so its idiomatic in typescript projects to use it even if you are not using npm

Ah, okay. should be done.

Zamiell:
awesome, thanks evan as always

PR #2701, opened 2022-11-01
holiday mix

vEnhance:
- Fix path in update_variant_files.sh
- Holiday mix
- Valentine mix


Zamiell:
should we save these variants for when the actual respective holidays arrive?

vEnhance:
I don't mind either way.

Zamiell:
ok, do you want to split out the bugfix part of the PR for now?

vEnhance:
OK split off in #2702

vEnhance:
> should we save these variants for when the actual respective holidays arrive?

@Zamiell there seems to be a bit of interest in Holiday Mix competition in the Hanabi Central server
Maybe we can merge this in a week-ish so people have time to practice and play the comp seeds over late Nov / early December?

Zamiell:
ok

Zamiell:
do you want to get rid of the valentine mix from the pr then

vEnhance:
> do you want to get rid of the valentine mix from the pr then

I could but I think it's more trouble than it's worth since variant ID conflicts could potentially arise between now and February

Zamiell:
I feel like we should probably wait until February to add the valentine's variant though right?

vEnhance:
I guess so
I'll just revert the Feb thing and reapply it on a later patch, it's only a couple lines

vEnhance:
Ready to merge whenever then I'll poke libster about making comp happen

Zamiell:
how about we merge it on dec 1

vEnhance:
sure, close enough

vEnhance:
> how about we merge it on dec 1

happy december :)

Zamiell:
its live now, thanks!

shemetz:
Why is this new variant called "Holiday Mix" when, as far as I can tell, it's just Christmas colors?

image

Zamiell:
what do you want to call it

shemetz:
Christmas Mix

PR #2706, opened 2022-11-08
Savespecnotes

Krycke:
Fixing the previous bugs and made some cleaner code.


Zamiell:
did you find out the bug that caused the crashing/deadlocks

Krycke:
Yeah, i found one thing that was visible in the logs, and i found another bug related to new sessions, that probably caused don't issues

Zamiell:
thanks

PR #2708, opened 2022-11-10
database setup edit

vEnhance:
Can someone check this?

I needed the latter to get the thing to work on a new device


Zamiell:
why does the hanabi user need the ability to create new databases? that doesn't sound right

vEnhance:
hmm I don't remember anymore --- I just had a note to myself somewhere that I had to do this to get it to work but I might have been hallucinating

I guess I'll just close unless it comes up again

PR #2712, opened 2022-11-13
Don't round down cluesStillUsable

### LINKED ISSUE #2519: Number of remaining possible clues rounded down in clue starved
In clue starved, if you hover over the efficiency and look at Remaining possible clues: n, it will round down n to the nearest integer. This is probably intended and it is logically correct, but it is slightly worse for reviews and for planning future lost clues.

vEnhance:
Closes #2519. I checked and the display seems fine if the number is an
integer (no trailing ".0") so I don't know why the Math.floor was there
to begin with, to be honest

(The Math.floor was introduced in 34ebfcfdf.)


vEnhance:
OK I understand the purpose of the floor now -_-

Zamiell:
add a comment so we know for net time

vEnhance:
OK

Zamiell:
thanks

PR #2714, opened 2022-11-14
fix: eff counter colour for null/dark null

chennisden:
They are no longer hard variants.


chennisden:
I don't know if dark null is still a hard variant or not (I assume not)

Zamiell:
thanks

PR #2715, opened 2022-11-14
Put connect messages in a go routine

Krycke:
Made the friendsHistoryID little sharper and put connecting Messages in a separate go routine to not hinder other connecting users if one is slow

### DIFF
--- a/server/src/models_games.go
+++ b/server/src/models_games.go
@@ -398,21 +398,21 @@ func (*Games) GetGameIDsFriends(
 	}
 
 	SQLString := `
-		SELECT DISTINCT games.id
-		FROM games
-			JOIN game_participants ON games.id = game_participants.game_id
+		SELECT DISTINCT game_participants.game_id
+		FROM game_participants
 		/*
 		 * We must use the ANY operator for matching an array of IDs:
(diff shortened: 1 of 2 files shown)


Zamiell:
can you describe the db changes in more detail

Krycke:
The sql in only stripping the two joins, as we are only interested in the games.id, which we already have in the participation table. The query itself is actually not changed.

Zamiell:
ok thanks

Zamiell:
future steps:

can we do a PR to lazily give friend data when they click on the friend history button?

future questions:

the above won't solve the root of the problem because as long as the DB query takes longer than 30 seconds the server will "deadlock", so are future DB call optimizations needed?

Krycke:
Agree, right now the query reads all database rows and go through them to find the ones that are played by friends, after that it does the same thing again to find the ones played by you, then take the diff of those. When all that is done we sort the set, and return only 10 rows. What we can do is having an index on when a table is created, and only go through the lines from games played the last week, or month or something. That will make this alot quicker.

But that said, by putting this on a separate go routine it shouldn't deadlock.

PR #2718, opened 2022-11-16
Patches a typo in CommandNote

Krycke:
Realized we are escaping the wrong message when storing a new note.

Also removes a leftover debug line


Zamiell:
thanks

Krycke:
This one

Krycke:
I would believe that the notes probably are getting escaped just before adding them to the database. Now that the code correctly escapes the notes before adding them to the server, the escaped note will be escaped again, rendering the text as shown in the picture.

PR #2719, opened 2022-11-16
Fixes trash bin size and position in suit is known

### LINKED ISSUE #2716: [Bug] The automatic trash bin don't jump up to the top left corner in all cases
trashbin

Krycke:
Closes #2716


Zamiell:
can we unify all the 0.2 constants to avoid violating DRY

Zamiell:
good thanks!

PR #2720, opened 2022-11-18
Removes Html escaping on the server

Krycke:
Removes html escpaing in the server, it's a job for the client

### DIFF
--- a/server/src/command_note.go
+++ b/server/src/command_note.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"html"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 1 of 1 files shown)


Zamiell:
it should be removed from the client instead

Krycke:
But it's not a server issue. If a bit for example is on the client side, we don't want to send html escaped notes to it. And also the notes written by the player is never passed to the server and back again, so the client needs to escape it's own text.

Zamiell:
> the notes written by the player is never passed to the server and back again

when you write a note, it is immediately transmitted to the server.
right?

Krycke:
It is, but it's not received and stored if the player is the one that sent
it.

Den fre 18 nov. 2022 16:16James ***@***.***> skrev:

Zamiell:
k

Zamiell:
i thought about forcing all notes to be sent to the server before being displayed, but that might cause noticeable lag when you write notes, which is undesirable.

currently, we should remark that the server does perform server-side sanitization of chat messages, but leaves sanitization of notes to the client.
i don't really like this asymmetry, as it seems confusing and hard to reason about, but we can leave it for now

PR #2724, opened 2022-11-28
Fixes empathy in hypos with morphed cards

Krycke:
If we know we have mophed cards that create an impossible deck, this will break empathy on all cards. We can remove them to make it work better.


Zamiell:
thanks

PR #2726, opened 2022-11-29
Morph updates

Krycke:
A few updates involving morphing
- Morph to blank or unmorph
- Fade out invalid cards in morph modal
- Use morphed cards when deducing
- Copy morphed cards with /copy
- Use escape to close the morph modal

### DIFF
--- a/packages/client/src/game/reducers/replayReducer.ts
+++ b/packages/client/src/game/reducers/replayReducer.ts
@@ -4,6 +4,7 @@ import { ensureAllCases } from "@hanabi/data";
 import produce, { Draft, original } from "immer";
 import { nullIfNegative } from "../../utils";
 import { ActionIncludingHypothetical, ReplayAction } from "../types/actions";
+import { CardIdentityType } from "../types/CardIdentityType";
 import GameMetadata from "../types/GameMetadata";
 import ReplayState from "../types/ReplayState";
 import gameStateReducer from "./gameStateReducer";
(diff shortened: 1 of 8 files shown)


Zamiell:
> Fade out invalid cards in morph modal

what does that mean

Krycke:
morph

It uses the empathy on the card you are morphing to help you morph to only valid cards. You can still click and morph to all cards, but then you know are creating an invalid state, that will probably break the empathy somewhere.

packages/client/src/modals.ts
```diff
@@ -140,13 +144,14 @@ export function askForMorph(
       variant,
       getMorphModalSelection(),
     );
-    if (cardIdentity.suitIndex === null || cardIdentity.rank === null) {
+    if (cardIdentity.suitIndex === -2 || cardIdentity.rank === -2) {
```
Zamiell, line 147: don't use magic numbers here, isn't there an enum for this

packages/client/src/modals.ts
```diff
@@ -501,3 +512,16 @@ function fillMorphModalWithRadios(
   });
   placeHolder.append(table);
 }
+
+function containsPossibility(
+  possibilities: ReadonlyArray<readonly [number, number]>,
+  suit: number,
+  rank: number,
+) {
+  for (const possibility of possibilities) {
```
Zamiell, line 521: shouldn't we use array.some?

Zamiell: whenever you have this pattern you generally want to refactor it to array.some

packages/client/src/game/reducers/replayReducer.ts
```diff
@@ -252,9 +253,13 @@ function hypoAction(
   // The morph action is handled here. Also take note of any draws that conflict with the known card
   // identities.
   if (action.type === "morph" || action.type === "draw") {
-    if (action.suitIndex === -1 && action.rank === -1) {
+    if (
+      // eslint-disable-next-line isaacscript/strict-enums
+      action.suitIndex === CardIdentityTypes.Original &&
```
Zamiell, line 258: the type of suitIndex should be `number | CardIdentityType` I think

packages/client/src/game/types/CardIdentityTypes.ts
```diff
@@ -0,0 +1,5 @@
+enum CardIdentityTypes {
```
Zamiell, line 1: it should be non-plural

packages/client/src/game/reducers/replayReducer.ts
Zamiell: because if you are disabling the rule it probably means you have something incorrectly typed

Krycke: Yeah, i tried that but it didn't work. And i didn't do anything differently in the other place. Tried to find information on Google but didn't find anything, and tried to read the documentation of that rule, but ended up on a 404 page. It works of course of i use "value of" but that's even uglier. I can look a little further

Zamiell: i wrote the rule, why is the documentation 404

Krycke: Don't know, might be wrongly linked in my workspace maybe?

Zamiell: what link are you clicking on

Zamiell:
nice, thank you!

Zamiell:
i got rid of the CardIdentityType enum, I thought of a better way to do this using a new "unmorph" action

PR #2728, opened 2022-11-30
Fixes drag problems in hypo

### LINKED ISSUE #2389: when in hypo, cards can have wrong orientation

Krycke:
Closes #2389

### DIFF
--- a/packages/client/src/game/ui/HanabiCardMouse.ts
+++ b/packages/client/src/game/ui/HanabiCardMouse.ts
@@ -312,6 +312,17 @@ function dragStart(card: HanabiCard) {
   // We need to change the cursor from the hand to the grabbing icon.
   card.setCursor();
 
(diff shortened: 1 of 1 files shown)


Zamiell:
oh, this makes hypos too hard to use unfortunately

Zamiell:
its very difficult to drag rotated cards to where they need to go

Krycke:
Hm... This also fixes the bug with shadows not getting removed when your not in keldon mode. Maybe we should add a keldon check around it. And just keep it for the few that uses keldon mode?

Zamiell:
i guess?

PR #2729, opened 2022-11-30
Added /untag

### LINKED ISSUE #2651: /untag should do the same than /tagdelete
to remove a tag "/untag [tag]" should work

Krycke:
Closes #2651


Zamiell:
looks like it fails ci

Zamiell:
do you not run prettier on save in vscode?

Issue #2180, opened 2021-05-04
playing a card in hypo of throw it in a hole breaks hypo until ended


timotree3:
Steps:

1. Go to https://hanab.live/shared-replay/520426#1
2. Blind play slot 4
3. Try making a move as Bob (you can't)
4. End the hypo. The blue 1 you blind-played is missing
5. Go forward and backward a few turns. The blue 1 has reappeared

Krycke:
This one is solved

PR #2735, opened 2022-12-07
Fixes ThrowItInAHole Hypos in replay and displays "Plays" label

Krycke:
Needed to know if the game was finished, to be able to "not throw it in a hole" in replay hypos


Krycke:
Closes #2180

Zamiell:
thanks!

PR #2737, opened 2022-12-08
Enable spectators notes to effect cards

### LINKED ISSUE #2736: Empathy should take notes from player into account, when shadowing
When you are shadowing Alice, a card should change according to Alice' notes. If she is writing [f][b3] the card should morph into b3 with a light blue border around.

Krycke:
Closes #2736


Zamiell:
thanks!

PR #2744, opened 2022-12-16
Red timer on player running out of time

### LINKED ISSUE #2669: Second indicator when <10 seconds
Hi,
I played one game today and I'm sure to have the sound not starting. And I think it's the third time it happen to me.
I was using chrome, and I think my turn started on a youtube firefox tab. Not sure what caused it.

But if you could add another indicator visible in game when you run out of time it would be great.
The first thing which come to my mind was a red background where the timer is.

Krycke:
Closes #2669


Zamiell:
thanks

PR #2745, opened 2022-12-16
Let anyone set leader

### LINKED ISSUE #2642: Give the lead when you double clicking the crown (and the leader of the game is away)
Initially I wanted only if the leader was away, but if all players don't abuse of it, that's even better in most cases.
Don't let spectators use that functionnality.

I guess even in important games, if someone abuse of it, he will just not be invited anymore next time, so that should work.

Krycke:
- Anyone can already use /setLeader chat command. Let's make it easier by letting anyone click the crown to change leader.

- Closes #2642


Zamiell:
ok

PR #2746, opened 2022-12-16
Disable unplayable warnings in hypothetical

### LINKED ISSUE #2656: Known Unplayable card warning can display incorrectly in Hypo after editing cards
In the example below, the multi 3 was originally a different card, but was edited in the hypo to be the multi 3.

The multi 3 should be playable, however, the `Are you sure you want to play this card? It is known to be unplayable based on the current information available to you. (e.g. positive clues, negative clues, cards seen, etc.)` warning appears unnecessarily.

image

Here are two possible suggestions for different ways to address this scenario:
Suggestion 1) Make the logic for this alert work correctly for edited cards
Suggestion 2) Turn off this alert entirely when in hypos

Krycke:
Closes #2656


Zamiell:
thanks

PR #2754, opened 2022-12-19
Hard variant -> 1.33

vEnhance:
Implements https://github.com/hanabi/hanabi.github.io/issues/1036


Zamiell:
can we define this as a constant somewhere instead of making the change multiple times

vEnhance:
Sure, where do you want to put it?

Zamiell:
some constants.ts file

vEnhance:
OK I put it in `client/src/constants.ts`

Zamiell:
good thanks

PR #2756, opened 2022-12-20
Server *.tmpl has some mixed tabs/spaces

vEnhance:
This changes them to all spaces


Zamiell:
Good PR thanks.

Can we tell prettier to format tmpl files as html?
That would solve this problem in CI, which is better

PR #2760, opened 2022-12-20
Prettier tmpl

vEnhance:
Adds https://www.npmjs.com/package/prettier-plugin-go-template
and then reformats all the Go templates


Zamiell:
perfect thank you

PR #2763, opened 2022-12-21
Update Playing relationship of players not present

### LINKED ISSUE #2758: You can get stuck "playing a game" even if it's in Shared-review
Sometimes you can get stuck after a game has finished and entered shared-review. You can't create a new game until the everyone in the shared replay has left and the game has been removed. Until then the server claims that "you are already playing a game".

Suggestion for a fix. When the server want to alert you that "you are already playing a game", check to see if that game is actually running, if not, remove the map entry, and release the player.

Krycke:
Players need to be deletes from playing tabke relationship even if they are not present at the table when it terminates.

Closes #2758

### DIFF
--- a/server/src/game_end.go
+++ b/server/src/game_end.go
@@ -365,6 +365,8 @@ func (t *Table) ConvertToSharedReplay(ctx context.Context, d *CommandData) {
 				logger.Info(p.Name + " was the owner of the game and they are offline; " +
 					"passing the leader to someone else.")
 			}
+			// We still need to update table relationships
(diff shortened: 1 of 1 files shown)


Zamiell:
nice bug fix thanks

Krycke:
Yeah, when testing my original suggested fix to the problem i actually stumbled upon the real bug. I don't think a sanity check if needed anymore in the tables relationship, but if we see more problems, that could still be a thing to do

Zamiell:
very nice work

PR #2767, opened 2022-12-22
Improve cspell configuration and scripts

timotree3:
- Make cspell in CI obey gitignore (it already did in VS Code), so we can remove some things from the ignored file list
- Switch from using python in the script (it defaults to python2 on my machine, which means that the script can't handle non-ascii json and crashes)
- Add support for comments in cspell.json to `check-orphaned-words.sh` (cspell already supports comments)


Zamiell:
thanks

PR #2769, opened 2022-12-22
Require 2 votes to kill a 2 player game

### LINKED ISSUE #2765: Vote to kill in 2p should require both players to vote
Currently in 2p, just one player voting to kill is enough to end the game. It would be more useful if it required both players to vote, since they can always double click to terminate if they want to kill dictatorially anyway. Related #2414

timotree3:
Fixes #2765


Zamiell:
thanks

Issue #1234, opened 2020-06-13
arrows are on the wrong side of the hand in keldon mode


Zamiell:
@robertos

Zamiell:
should be fixed in his upcoming PR

Zamiell:
seems to be fixed now

PR #2787, opened 2023-01-11
Keep the name of the shared replay

### LINKED ISSUE #2781: Shared replays should keep their names
Why are we changing the name of the replay to "Shared replay of Game #12345". Is it not nicer to keep the name, as is?

Krycke:
Creates the replay name:
My random name (Game #1234)

Closes #2781


Zamiell:
thx

PR #2790, opened 2023-01-11
Display morphed card on empathy

### LINKED ISSUE #2776: Show morphed hidden hand cards
When in hypothetical and not showing hidden cards. You can't get a morphed card to show (both when already in the hand and not yet drawn on the turn you start the hypo) in the hand of the player whose cards are hidden.

Krycke:
Closes #2776


Zamiell:
thx

PR #2802, opened 2023-01-20
Fixing the selector arrow in create new game

### LINKED ISSUE #2798: css is messed up on create game tooltip
image

the magnifying glass row of icons is misaligned, needs to be fixed somewhere in the CSS or HTML i guess

Krycke:
Closes #2798
... again ...

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -6,6 +6,8 @@
   "enableGlobDot": true,
   "useGitignore": true,
   "ignorePaths": [
+    //Ignore yourself
+    "cspell.json",
     ".git/**",
     // Assets which are not edited by hand
     "*.cdr",
--- a/cspell.json
+++ b/cspell.json
@@ -20,6 +20,8 @@
     // Minified dependencies
     "*.min.js",
     "*.min.css",
+    //The css might have strange paths
+    "hanabi.css",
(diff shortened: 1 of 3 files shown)


Zamiell:
what is changing in alpha.css? that file is supposed to remain vanilla

Krycke:
Yeah i thought so as well, but that arrow on the default variants selector that is grey and looks like it's deactivated, is a background picture that is defined with a color in alpha. I could override the value in hanabi.css if that's better.

Zamiell:
Yeah, make a new section with a comment called "alpha.css tweaks" and explain why and what you are doing

Zamiell:
looks like it fails ci

Krycke:
It failed spellcheck because the css has the following line:

`... vg' width='40' height='40' preserveAspectRatio='none' viewBox='0 0 40 40'%3E%3Cpath d='M9.4,12.3l10.4,10.4l10.4-10.4c0.2-0.2,0.5-0.4,0.9- ...`

And the spell check is failing from the `Cpath` part.

How do you want this to be solved?

Zamiell:
lets just add an ignore for the file to cspell.json

Zamiell:
thanks

PR #2809, opened 2023-02-01
Seed stats

### LINKED ISSUE #1057: Improved player/game statistics
Not sure if this has already been discussed before but I think it'd be cool to see some additional game statistics functionality, mostly to get a sense of how "hard" a deal is.
In "other games":
- sort table by score
- strikeout rate
- non-strikeout average score or histogram
- percentile rank for games you participated in

In profile:
- some kind of player percentile ranking per variant/player# combo. This could either just be max score or the aggregation of all percentile rankings for each game. 

I'm happy to contribute technically / submit a PR for this, just want to make sure it's something the community wants and Zamiell/other maintainers think is acceptable from a server load point of view (I don't think it should be too crazy, we can probably start with calculating stats on page request?)

### LINKED ISSUE #1759: Leaderboard for every Variant
It would be nice to have a leaderboard for every variant to see the team that currently has a max score for that variant.   

image

For example, here you can see that there are a total of 10 perfect scores, but none of them are listed on the front page as it only shows the most current games.  

It would be nice to see those max scores to have a quick reference.  Where you can sort by player count, date and other stats. 

https://github.com/Zamiell/hanabi-live/issues/1092

### LINKED ISSUE #2257: Add score histograms to seed history
Hi!

I'd like to add a histogram of scores to the seed history page.

I figure I can calculate them here: https://github.com/Zamiell/hanabi-live/blob/master/server/src/command_history_get_seed.go

Any objections or suggestions before I get cracking?

vEnhance:
Closes #2257 
Closes #1759 
Closes #1057 

Implements the following changes:

Adds a "View seed data" from the client view that takes you to the seed page
image

Sorts by descending score by default now, rather than by game ID

Adds a dumb bar chart to the seed page

image

It's possible this looks extremely dorky on production if the 0 bar ends up being way way taller than the others or something; I don't know what the real-world data looks like. If it does, then we might consider removing the 0 bar and then recording the fraction of games with score 0 in a different way.

This uses `chart.js` to draw the chart. That's sort of a big dependency. I'd certainly be open to ways to use a less heavy import.


packages/client/src/lobby/history.ts
```diff
@@ -319,16 +323,16 @@ export function drawOtherScores(
 
     // Column 6 - Seed. Chop off the prefix.
     const match = /p\dv\d+s(\d+)/.exec(gameData.seed);
-    let seed: string;
+    let seedNumber: string;
```
Zamiell, line 326: shouldn't this be called `seedString`

Zamiell: oh, i see, because it is a subset that is a number.
can we call it `seedNumberSuffix` or something like that?

Zamiell:
how big is the dep

vEnhance:
> how big is the dep

 200 kb

Zamiell:
even when minified? that seems pretty big

the current JS is 1.3 meg, so that would make things go to 1.5 meg

Zamiell:
15.3846% increase

Zamiell:
oh i guess it would only apply when not on the main site and on the specific seed pages, right?

vEnhance:
> oh i guess it would only apply when not on the main site and on the specific seed pages, right?

yeah

Zamiell:
oh, then we don't care about the dep then

Zamiell:
cool PR, i'll reboot the server now

PR #2811, opened 2023-02-02
Closes #1798

### LINKED ISSUE #1798: Morph cards in duck
[yellow] or [5] notes don't work currently because the "blank" image takes precedence over everything.

vEnhance:
Show card backs in duck (and cow&pig)


Zamiell:
nice bugfix

PR #2822, opened 2023-02-10
Preserve password on remake table

### LINKED ISSUE #2683: Feature requests for remake table
- [ ] [Manage specs](https://discord.com/channels/140016142600241152/333680903748321282/1027655997164113960)
- [ ] [Preserve game password](https://discord.com/channels/140016142600241152/733076625079402647/1027705966503919697)

(@Zamiell you can assign to me)

vEnhance:
Closes #2683

I decided not to implement the thing that makes spectators join as
players because of Krycke's comment in #2683. (Indeed, seems wrong to
me in at least one case: if I'm say playing a game with some friends
that's pw-protected, and I restart the game, the random specs watching
that didn't enter a password probably shouldn't be joined.) So I'm gonna
say someone should open a new issue describing what the "correct"
behavior should be.

### DIFF
--- a/server/src/command.go
+++ b/server/src/command.go
@@ -83,6 +83,9 @@ type CommandData struct {
 	// (e.g. the mutex lock is already acquired and does not need to be acquired again)
 	NoTableLock  bool `json:"-"` // To avoid "t.Lock()"
 	NoTablesLock bool `json:"-"` // To avoid "tables.Lock()"
+	// More remake table shenanigans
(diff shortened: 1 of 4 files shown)


Zamiell:
thanks

Zamiell:
can you confirm what the behavior is now for specs?

vEnhance:
> can you confirm what the behavior is now for specs?

same as the old behavior, which i think means they join as specs for restart and are booted for a remake

Issue #2831, opened 2023-05-14
Broken database entries for some games


kesslermaximilian:
While downloading games via the export api, I noticed that there are some games without a list of players being returned in the exported json.

Specifically, for 'No Variant', these games are
- https://hanab.live/export/357913
- https://hanab.live/export/357915
- https://hanab.live/export/358627
- https://hanab.live/export/358630
- https://hanab.live/export/358632
All of these show an empty list of players as well as actions (the latter not being a problem).
I also think that there is at least some occurrences also for other variants, but I have not investigated that so far.

This leads to unexpected behaviour at different parts of the website:
- When trying to watch a replay of these games (try e.g. https://hanab.live/replay/357913), you will receive an error message stating that the game could not be constructed (which makes sense, since in particular the number of players is unknown)
- When listing played games via the api, pages that would contain the above games return less items than expected, concretely
    - https://hanab.live/api/v1/variants/0?col[0]=0&page=1633&size=100 returns only 98 rows of data, the first two games listed above would be contained in that range
    - https://hanab.live/api/v1/variants/0?col[0]=0&page=1635&size=100 returns only 97 rows of data, the last three games listed above would be contained in that range

As of my knowledge, this is not big of an issue, these just seem to be some broken entries in the database and I don't see a particular need to do something about it, although it would of course be nice to resolve the issue by fixing/deleting the entries.

Also it would of course be interesting to know what caused these invalid entries to be in the DB in the first place, but I have no idea regarding this.

Aigul9:
@kesslermaximilian There are plenty of games missing (around 6k at the moment). i think one of the main reasons is that unnecessary/similar variants were deleted and replaced by the new ones, therefore there was no need to keep junk games in the db.

kesslermaximilian:
I know that there are a lot of 'missing' games in the sense that not all gameIDs ≤ (current max) are ids of actual games, but I do not consider that a problem.
If you query these, you will get an error stating 'Game #<id> does not exist in the database', which I think is perfectly fine and self-explanatory.

The games I mentioned above, however, fail this usual pattern, since there are actual entries in the DB for them, but the data for them is incomplete (missing player names), which leads to this ' Failed to initialize the game. Please contact an administrator.' error, which is a very different kind of error.

So I think that either, player names for these games should be re-added (if possible) or the games should be deleted entirely to ensure consistency of the stored data.

kesslermaximilian:
[...]

PR #2832, opened 2023-05-23
Add check on number of players on game api export

kesslermaximilian:
Related to issue #2831:
Export function currently reported incomplete player lists when using the export API, resulting in invalid output.
This introduces a check on the correct number of players, analogously to the check in command_replay_create.go when loading full replays from the DB.

Note that this of course does not prevent inconsistent player lists from occurring, they are just detected properly now.

### DIFF
--- a/server/src/http_export.go
+++ b/server/src/http_export.go
@@ -79,6 +79,20 @@ func httpExport(c *gin.Context) {
 		options = v
 	}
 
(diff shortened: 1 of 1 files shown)


Zamiell:
thanks

kesslermaximilian:
happy to help

PR #2839, opened 2023-07-07
DB: Delete games upon participant deletion

kesslermaximilian:
This adds a trigger to the `game_participants` table so that whenever a participant of a game is deleted
(this already automatically happens whenever a user is deleted), the corresponding game is deleted as well.

Some details:
Implementation is the same as https://stackoverflow.com/questions/30281819/postgresql-delete-trigger. I suppose because the code snippet is so short, this does not cause copyright issues. If you think so, we could e.g. put a link into the source code to StackOverflow, but I don't think it's needed.

Also, not that we could e.g. add more checks such that also after each update or insert to the game_participants table (or the games table even), we check that there is the correct number of participants in the database, but this would be more work (I currently don't know how to do it properly) and is likely not needed: I don't think that we ever insert a game into the DB without inserting its participants or just randomly add participants, unless of course there is other bugs.

This PR should fix that manual deletion of users can break the database consistency (#2831), which is probably all we want for now.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -210,7 +210,7 @@
     "pidfile",
     "pingable",
     "playerinfo",
-     "plpgsql",
+    "plpgsql",
     "pnpm",
     "pquestion",
     "prasmussen",
--- a/cspell.json
+++ b/cspell.json
@@ -210,6 +210,7 @@
     "pidfile",
     "pingable",
     "playerinfo",
+     "plpgsql",
     "pnpm",
     "pquestion",
     "prasmussen",
--- a/install/database_schema.sql
+++ b/install/database_schema.sql
@@ -167,6 +167,17 @@ CREATE TABLE game_participants (
     CONSTRAINT game_participants_unique UNIQUE (game_id, user_id)
 );
 
+CREATE FUNCTION delete_game_of_deleted_participant() RETURNS TRIGGER AS $_$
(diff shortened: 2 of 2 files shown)


kesslermaximilian:
Since probably you don't want to re-initialize the database, you'll have to add these triggers to the DB with a prompt manually once.

kesslermaximilian:
Stuff that I tested with a local installation of the server:

- Deletion of a user correctly cascades to deleting all entries from the `game_participants` table as well as all games from the `games` table, works with single or multiple games
- Manual deletion of games does not cause problems, the game and its participants are deleted (same behaviour as before)
- Manual deletion of entries in the `game_participants` table also works, deleting the corresponding game which in turn (potentially) deletes more from the `game_participants` table (namely the other participant entries for the same game)

As this is not an easy change, I suggest that you test this locally as well beforehand (maybe you have a copy of the server running or sth) and not put it directly into production.

kesslermaximilian:
Should I fix that spell check 'issue' by explicitly disabling it for that line?

Zamiell:
add `plpgsql` to allow words in cspell.json i think

Zamiell:
very nice PR, thank you

kesslermaximilian:
you're welcome, happy to help!

Issue #2115, opened 2021-03-02
Add new variant: sudoku


newduke:
Sudoku variant.

Each stack must start at a different rank (other plays are bombs)
Each subsequent play must be the next (cyclic) rank (eg, 2..3..4..5..1)
Stack is complete when it has all ranks.

Each suit has 2 of each rank because the variant is symmetric by rank.

cjwomack:
Sounds interesting... but what would be sudoku variant conventions? So no saving of 2 or 5s... And would the starting convention to try to clue two cards eg... r2 and r3 are next to each other in someones hand so clue red?

timotree3:
On turn 1 you could simply blind play a slot of your choosing :laughing:

jakestiles:
This seems fun!

If this is too easy, perhaps it could be changed so each color has a rank that it is forced to start at. Leftmost color starts at 1, next color starts at 2, etc.

newduke:
It's quite easy to force each suit to start at its rank, but "sudoku" (the original) doesn't have that restraint and I think it will be interesting to see how players use that to their advantage. 

I suspect that this variant will be a bit easy, but we can add several dark suits or other peculiarities to spice it up.

Romain672:
To make it easier to play I got this idea:
 - r0, y1, g2, b3, p4 are initially played on stacks
 - yellow suit goes from 2 to 6
 - green suit goes from 3 to 7
 - blue suit goes from 4 to 8
 - purple suit goes from 5 to 9
 - 1 clue touch 1&6
 - 2 clue touch 2&7
 - 3 clue touch 3&8
 - 4 clue touch 4&9

This is similar than the first post, buy is maybe more easy to play.

kesslermaximilian:
We should close this now.

PR #2847, opened 2023-07-24
Add Sudoku variant

kesslermaximilian:
This implements #2115 and is based on newdukes work at #2114. The first commit here is just a rebased version of newdukes commit from some 2 years ago.

Some details on the features:
- UI shows the starting value of each stack once it is determined (either when a card has been played or if it is the last remaining stack without a card played)
- UI shows number of cards played for each stack, since one loses track of the individual progress quite quick in this variant
- Empathy (for trash) and maximum score calculation adjusted to Sudoku variants

Some implementational details:
- Server games now have an attribute 'playStackStarts' which stores the starting values of each stack. It is only used in Sudoku variants (otherwise it is never read)
- Server also has corresponding implementation to calculate maximum scores in Sudoku variant
- Added some helper methods in `misc.go` to compute minima and check arrays for contained elements
- Client also has new `playStackStarts` attribute in the `GameState`, analog to Server
- Client UI has a second row of LabelTexts below stacks that are only created in Sudoku variants (to write the number of played cards in there)
    - Creation of this row is toggled by the new `showStackStarts` attribute in the `variants.json` file

In case you would like more documentation in the code, I could add more detailed documentation to the changes.


Zamiell:
Make sure it passes ci

kesslermaximilian:
Yes, I'm on it currently. Somehow the linters do not work properly for me locally

kesslermaximilian:
Passing now

Zamiell:
did you test the new ui elements in keldon mode

kesslermaximilian:
Oh no, totally not, I forgot that existed. Will do

kesslermaximilian:
Works

kesslermaximilian:
Ah no, when it reaches your turn, it overlaps a little, let me see what we can do there

Zamiell:
i dont think we have room for a new ui element, is there some other solution we can do

kesslermaximilian:
An alternative would be to put the second line in the Hover above the Stacks. But I sort of don't like this, because in contrast to how suits behave, that information changes all the time, so you would like to see it.

kesslermaximilian:
I'm very much a fan of these, at least in Non-Keldon mode. As a compromise, we could put them in the Hover in Keldon mode and leave them as implemented right now otherwise.

Zamiell:
ok

kesslermaximilian:
One suggestion we had from Tony was the following: Say the stack starts at 2 and 234 have been played already. Then we could put <bold>234</bold>51 under the stack. This fits in one row and shows all the information needed. Even has the additional upside that you can see the last card of each suit etc. What do you think of that?

Zamiell:
seems fine

kesslermaximilian:
As I see it, the Konva.text does not support different font styles within the same text element. Do you know how to implement above approach? I forgot to mention that the idea in the above example would be to have 234 in bold and 51 not in bold, to also see how many cards have been played.
But in case we can't do the boldface properly, I think I will go with the hover solution in Keldon mode

Zamiell:
no idea, you could just make two different text elements

Zamiell:
konva sucks its ancient technology

kesslermaximilian:
> no idea, you could just make two different text elements

Problem with two different elements is that it's hard to align then: Since the number of boldface characters is varying, we would either dynamically need to change sizes of these fields (and calculate these so that in total, stuff is centered), or have two at the same position and somehow align text inside these properly). I don't feel comfortable getting this to work in a consistent way (bold font is probably bigger etc), so in case there is no easier way, I will not implement it like that.

In case someone has further suggestions, I'm open for them, but probably it'll be the hover solution in Keldon.

kesslermaximilian:
> konva sucks its ancient technology

Yeah I see

Zamiell:
go is pretty good, it has some downsides, but nowadays i prefer typescript because of all of the tooling involved (prettier+eslint is amazing, as well as package managers, monorepo tools, etc.)

kesslermaximilian:
UI is now also done for Keldon mode (only >2 players, since in 2p, there was still space) and shown in the Hover there. Any more things that came to your mind before we can merge this?

PR #2850, opened 2023-07-26
Add more Sudoku variants

kesslermaximilian:
Since Sudoku works, I think we should add these to the server as well. Have also seen people requesting them.


kesslermaximilian:
Thanks!

PR #2854, opened 2023-07-27
Rework Sudoku UI

### LINKED ISSUE #2852: Sudoku UI
There have been some discussions about this already, and I haven't found a clean solution yet, so I'll just start an issue to let people discuss how the Sudoku UI should be.
I think we can agree that the UI should
- Make the stack starts clear. This is needed to figure out where the other stacks can be started at.
- Make it clear how many cards have been played already per stack. This is useful for priority and also just to track (endgame)-progress.
- Make it clear which 'set' of cards is played already. This is useful to reduce possibilities of clued cards in the own hand by good touch.

The current UI achieves the first two, but I hadn't thought of the third one, though admittedly, it'd really be something nice to have.
Suggestions to improve this that I heard so far were

- Say a suit starts with 2, then we can display 23451 in that order. We can then separate the played cards from the rest by an underscore, so if 23 are played, we can show 23_451
- Same, but the played cards are in boldface. Disclaimer: I currently don't know how to do this, since the Konva text-element does not support different styles within the same text element. Happy with help here.
- Remove 'x cards played' and replace it with a list of the played cards, e.g. '23 played'. Maybe with commas?

Further suggestions very welcome

kesslermaximilian:
Fixes #2852.

In particular, gets rid of all the UI code related to the second label row below the stacks, since we don't need these anymore. Also simplifies logic, since we don't have to do anything special in Keldon mode.

### DIFF
--- a/packages/client/src/game/reducers/gameStateReducer.ts
+++ b/packages/client/src/game/reducers/gameStateReducer.ts
@@ -8,7 +8,6 @@ import * as clueTokensRules from "../rules/clueTokens";
 import * as deckRules from "../rules/deck";
 import * as handRules from "../rules/hand";
 import * as playStacksRules from "../rules/playStacks";
-import { fillInRemainingStackStartIfUnique } from "../rules/playStacks";
 import * as textRules from "../rules/text";
 import * as variantRules from "../rules/variant";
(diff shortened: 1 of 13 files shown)


Zamiell:
nice thanks

PR #2857, opened 2023-07-27
Fix empathy up or down

### LINKED ISSUE #2856: Broken Empathy in Up or Down variant
The empathy in Up or Down is currently broken in some rare cases:
If both the start card and one of 1 or 5 is lost in a suit and nothing is played, the client does not recognize that even though the stack does not have a determined stack direction, it is still implictly determined. It will evaluate other cards as if we still had the opportunity to start the stack from both sides.

Example: Say no red card is played, but we lost 1, start and both 4s. Then 2 and 3 are trash, because we have to start with 5 now. Client still shows both 2 and 3 as useful (or even critical if unique) if they appear in players hands.

I guess it's such a rare situation that in practice, it does not really matter (people perform better than losing 1,start,4,4 or similar in one suit), but still a bug I'd say.

kesslermaximilian:
Fixes #2856. I hope this is correct now, could use some more testing for edge cases, but worked fine so far.

### DIFF
--- a/packages/client/src/game/rules/playStacks.ts
+++ b/packages/client/src/game/rules/playStacks.ts
@@ -9,7 +9,7 @@ import { CardState } from "../types/CardState";
 import { StackDirection } from "../types/StackDirection";
 import * as variantRules from "./variant";
 
-export function lastPlayedRank(
+function lastPlayedRank(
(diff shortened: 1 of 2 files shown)


Zamiell:
thanks

Issue #2859, opened 2023-07-27
sudoku ui


Zamiell:
![image](https://github.com/Hanabi-Live/hanabi-live/assets/5511220/be39d514-1008-47e8-af8e-8156a326ff82)

start text overlaps

suggestion: change `Start: [12345]` to `Start: [Any]`

Zamiell:
even after this suggestion, text is too cramped:

![image](https://github.com/Hanabi-Live/hanabi-live/assets/5511220/407dc9ee-356e-4743-bc98-db03f0e775c6)

Zamiell:
Also the labels don't look centered.
For example, look at the label under the green 5, the left bracket is to the left of the card, but the right bracket is aligned with the right of the card.

Zamiell:
![image](https://github.com/Hanabi-Live/hanabi-live/assets/5511220/235d61f5-d003-449f-82eb-0d6a9afde10f)

says 0 when a 1 is played under purple stacks ??

kesslermaximilian:
> ![image](https://user-images.githubusercontent.com/5511220/256542154-be39d514-1008-47e8-af8e-8156a326ff82.png)
> 
> start text overlaps
> 
> suggestion: change `Start: [12345]` to `Start: [Any]`

seems good, you already did that

kesslermaximilian:
> even after this suggestion, text might be too cramped:
> 
> ![image](https://user-images.githubusercontent.com/5511220/256544022-407dc9ee-356e-4743-bc98-db03f0e775c6.png)

For me it's way less cramped, probably depends on size of display? I can test some different solutions.

Zamiell:
> probably depends on size of display?

im on a super-normal monitor in 1920x1080 resolution

kesslermaximilian:
> ![image](https://user-images.githubusercontent.com/5511220/256544903-235d61f5-d003-449f-82eb-0d6a9afde10f.png)
> 
> says 0 when a 1 is played under purple stacks ??

I have no idea how this can happen. If you look at the implementation, the number computed is computed as (rankOffset + stackStart -1)%5 +1, so this being 0 would mean rankOffset+stackStart = 0, but offSet is nonnegative since its an array index and stackStart as well, since it's the rank of some card.

Also never saw this before.

kesslermaximilian:
> Also the labels don't look centered. For example, look at the label under the green 5, the left bracket is to the left of the card, but the right bracket is aligned with the right of the card.

[...]

PR #2860, opened 2023-07-28
Client: Sudoku: Fix UI bug showing 0 as stack start

kesslermaximilian:
Client sometimes showed 0 as stack start immediately after the first card of a stack played. Probably this is due to visibleRank not being updated properly before drawing the UI (the bug seems to occur a bit random), but getCardIdentity does not have this issue.

See also #2859


Zamiell:
Thank you!

PR #2862, opened 2023-07-30
Sudoku 4 suits

### LINKED ISSUE #2861: Sudoku with 4 Suits
I thought about it (and talked a bit with Hoddie), and I think we can make sense of 4-suit variants of Sudoku. The idea would be that we only have ranks 1,...,4 for each suit and then wrap around from 4 to 1.
The deck gets pretty small then (only 4 * 4 * 2 = 32 cards), so the pace/efficiency values would be:

- 3p: pace 4, efficiency 1.06
- 4p: pace 4, efficiency 1.14
- 5p: pace 1, efficiency 1.45
- 6p: pace 4, efficiency 1.14

(All assuming no dark suits, dark suits are just not winnable)

While 5p is certainly challenging, I think it's an interesting idea similar to 3-suit variants.

Would you accept a PR for these?

kesslermaximilian:
Fixes #2861.
Adds 4-suit variants of Sudoku as discussed there.


packages/client/src/game/ui/reactive/view/cardLayoutView.ts
```diff
@@ -132,7 +132,13 @@ export function onPlayStacksChanged(
 
   if (variantRules.isSudoku(globals.variant)) {
     // First, we will find out all available stack starts.
-    const availableStackStartsFlags: boolean[] = [true, true, true, true, true];
+    const availableStackStartsFlags: boolean[] = new Array<boolean>(
+      globals.variant.ranks.length,
+    );
+    console.log(availableStackStartsFlags);
```
Zamiell, line 138: .

kesslermaximilian: Oh yeah, that should not be there.

kesslermaximilian: it's removed now.

Zamiell:
I'll merge this on August 30th or so

Zamiell:
can you rebase this on main

kesslermaximilian:
In main, you now have two identical functions `initArray` and `newArray`, I think we should remove one.

kesslermaximilian:
Regarding the remaining linter errors here, I don't know what exactly those are and I think they're not caused by this PR...?

Zamiell:
Oh yeah I'll get rid of initarray in favor of newarray

Zamiell:
If lint passes on master then they have to be because of your pr

Zamiell:
if you look at the pr you can see you are removing eslint disable comments in the files changed tab

Zamiell:
if you rebase the pr on main again maybe the errors will go away

kesslermaximilian:
I'll try to do so the next days, not sure when exactly I'll have time.

PR #2895, opened 2023-12-11
Clean up sudoku

kesslermaximilian:
This cleans up some of the Sudoku code as requested:
- Removes null-assertion operator, instead uses a combination of `assertDefined` and more specific types to ensure safe index access.
- Clarifies the suitMaxScores variable, which was confusing before
- Fix a bug in the sudoku score calculation:
    - If both copies of a 5 were discarded, this caused wrong calculations on the maximum possible score.
    - This bug was introduced 1bd3fc87, since previously, the iterated range might have been empty when `lastDead +1 > 5` (which happens when both copies of a 5 have been discarded, but the `iRange` macro that had been introduced returns a reverse range in this case.
    - In general, I think that this usage behaviour feels very dangeruous to me. While it is documented, i would expect something like `eRange(5,2)` to just be empty, since this is also the standard behaviour in languages like C++, Python, etc. As you have seen, you (James) yourself have misused the macro yielding to above bug.
    - In case we don't explicitly need this reverse range behaviour, I would suggest removing it. Otherwise, we should probably check all refactored such range expressions and whether similar bugs have been introduced elsewhere.

### DIFF
--- a/packages/client/src/game/rules/variants/sudoku.ts
+++ b/packages/client/src/game/rules/variants/sudoku.ts
@@ -20,7 +20,8 @@ export function sudokuCanStillBePlayed(
   variant: Variant,
 ): boolean {
   const allDiscardedSet = getAllDiscardedSet(variant, deck, suitIndex);
-  const { maxScoresForEachStartingValueOfSuit } = sudokuWalkUpAll(allDiscardedSet);
+  const { maxScoresForEachStartingValueOfSuit } =
+    sudokuWalkUpAll(allDiscardedSet);
 
   const playStackStart = playStackStarts[suitIndex];
   assertDefined(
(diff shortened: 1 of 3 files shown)


Zamiell:
> If both copies of a 5 were discarded, this caused wrong calculations on the maximum possible score.

can you add a test please to prevent further regressions?

Zamiell:
i see that ci is failing because of prettier.
are you using vscode? it should be automatically summoning prettier as soon as you save any file, if it isn't then we need to troubleshoot your vscode setup because something is wrong on your end, you likely need to do e.g. `yarn install` at the root of the repo to install the deps, or the problem might be that you are opening individual files instead of using "open folder"

Zamiell:
> In general, I think that this usage behaviour feels very dangeruous to me. While it is documented, i would expect something like eRange(5,2) to just be empty, since this is also the standard behaviour in languages like C++, Python, etc. As you have seen, you (James) yourself have misused the macro yielding to above bug.

ok, I will refactor this function to remove the reverse functionality

kesslermaximilian:
No, I'm not using vscode, I run prettier only manually.

Zamiell:
what ide do you use?

kesslermaximilian:
webstorm

packages/client/src/game/rules/variants/sudoku.ts
```diff
@@ -163,30 +214,46 @@ export function getMaxScorePerStack(
...
+    undefined,
+    undefined,
+    undefined,
+  ];
+
+  // A map (index of stackStart) -> bool, denoting wether some suit is currently assigned to this
+  // stack start.
+  const assigned: Tuple<boolean, 5> = [false, false, false, false, false];
+
+  while (true) {
```
kesslermaximilian, line 235: I'm unsure on how to fix this linter error.
the localSuitIndex is incremented/decremented during the loop, and we want to stop the loop once it would be decremented to -1 (see line 349).

kesslermaximilian:
The mentioned tests are added now

Issue #2890, opened 2023-11-19
Feat: Randomized / Prefixed seed selection per account. OR: account linking for seed selection.


kesslermaximilian:
**Is your feature request related to a problem? Please describe.**
Quite some people have second accounts on this website for various reasons:
- Different accounts on different devices
- Keep track of separate stats for some project
- The Hanabi League
- ...

Since the current seed selection works in a way that the lowest seed number is chosen such that no team member has played that seed, this leads to newly created accounts (especially for things like the league) playing the same seeds:
- If you have not played the variant very often, you will very quickly end up playing the same seeds as on your main account. This is in particular a problem if you want to have one account for practicing (say, your main) and one for doing stuff like the league.
- If there is lots of different people creating new accounts for something like the league, they all play the same seeds. This discourages watching other games, as you likely will play the same seed.

**Describe the solution you'd like**
Some server-side setting that you can enable. I could imagine the following:
- Randomized seed selection: Instead of choosing the lowest seed number that is possible for a team, pick a random one. To not bloat the number of seeds, we should probably pick random out of all existing seeds, but with some specified minimum (say, 100 seeds). Then there's only 100 more seeds per variant in the database than strictly needed at all times (seems ok to me). Also, since a seed is just what we feed into the random number generator, this isn't a real storage problem anyway. This could either be enabled globally for the server, or we have some option to trigger it
- Seed prefixes: We could add an option to your account to prefix all selected seeds with some custom string, essentially creating a new namespace. Therein, we could work with the old implementation, i.e. choose mininum numbers that we append to these prefixes. We can read this setting from the table leader upon game start.
- Account linking: Add the option to 'link' accounts, which has the effect that if A and B are linked, neither account will play a seed that the other already has played. This is one of the cleanest in theory, but also the hardest to implement.

**Describe alternatives you've considered**
To avoid the specified problem without any server-side soltuion:
- Don't use alt accounts.
- Use fixed seeds upon game creation. I don't think this is feasible for stuff like the league, people *will* forget or play the same seed twice.

**Additional context**
This is really mostly interesting for the league, where we're working on S1 currently.

So the question is now: Would you be fine with one of these solutions to be implemented? This would of course mean one more setting per account that we have to store in the database.

pianoblook:
"Use fixed seeds upon game creation. I don't think this is feasible for stuff like the league, people will forget or play the same seed twice."

[...]

PR #2899, opened 2023-12-12
Account linking

kesslermaximilian:
This is a proposal implementation of #2890.

This introduces
- Commands `/link`, `/unlink`, `/links` that work in the same way as the `/(un)friend(s)` command to add *linked accounts* to a user.
- Server-side seed selection to prevent people from playing seeds that either they themselves or any of their linked accounts have played.

Regarding implementation:
- The implementation logic for storing the account is actually a bit simpler than with friends, since there is no need to store the reverse relationship, but otherwise analogue (also for the webclient).
- When selection a new seed, there is now the function `getNextAvailableSeed`, which will select the next possible seed for a set of players. It makes use of the `UserLinkages.GetBlockedSeeds` function, which is a single query t(but essentially 2) to retrieve all the seeds that cannot be used for the specified player set.

If this solution seems to complex, we can get rid of the account linking code and just modify the `getNextAvailableSeed` function to do some random selections.


Zamiell:
looks like this PR is actually pretty simple, so lets merge it, thanks

Zamiell:
can you document this feature though?

kesslermaximilian:
> looks like this PR is actually pretty simple, so lets merge it, thanks

Thanks!

PR #2904, opened 2023-12-12
rename command: links -> linked

kesslermaximilian:
People pointed out to me that this is probably the better naming. I'd suggest to keep /links as a (client-side) alias, so we'll just have both.


Zamiell:
sounds good

PR #2905, opened 2023-12-12
Bugix pm command: pass correct message

kesslermaximilian:
When refactoring this was broken, doing `/pm Alice hi` results in `Alice hi` being written to Alice.


Zamiell:
oops

PR #2906, opened 2023-12-13
Bugfix: Total number of cards in Up or Down + Dark Suit

kesslermaximilian:
Now that there is dark suits combined with Up or Down, we need to adapt the implementation to it, since there is now 6 cards of a dark suit in such a variant.

Since allowing this variant wasn't something I thought about a lot, I didn't think of it before. Also, there's nothing else I know about right now that we need to adapt about the rest of the code.


Zamiell:
sounds good

NZadeh:
Thanks for fixing this, we were just talking about this earlier today!

I'm wondering if the "real fix" should be removing the dark "Start" card. We were discussing earlier that this is more in the spirit of "dark" (but UD allows you to start either way, with 5 or 1, so it's actually more forgiving :)

PR #2911, opened 2023-12-14
server: add stackSize to variants, refactoring

kesslermaximilian:
Turns out the server did not have a sense of different stackSizes and therefore scores thus far, which is in particular why the stats page was bugged and it showed 20 for Sudoku 4 suits.

Variants already had a property `maxScore`, but it was not used consistently, I refactored (hopefully) all the usages to use `variant.maxScore` now when appropriate and set this value accordingly at initialization. Same goes for `stackSize`, which is added now as an additional attribute and used in the Sudoku code for more clarity.

I did a test game locally and it shows Sudoku correctly now on the stats page, the checkmark also works.


Zamiell:
nice thanks

PR #2914, opened 2023-12-15
server: Rework and fix efficiency calculations

### LINKED ISSUE #2913: Variant efficiencies are wrong on the (shared)-missing-scores page
Visit https://hanab.live/shared-missing-scores/Alice/Bob, the variant efficiencies for e.g. Sudoku 4 Suits are shown incorrectly.

Explanation:

I looked at the code in the `variantslogic/variant.go` file, apparently this is used to calculate efficiencies of variants for the missing-scores pages, so these are calculated server-side. But there's several things off:
- The file has its own list of variants, but we already have such a global variable in `src/variants.go`. Both are initialized in the `main.go`
- The file replicates lots of the methods of variants, mostly stuff like `isCriticalFours`. So in total, there is just two methods `isCriticalFours` and `IsCriticalFours`, with identical implementations.
- The efficiency calculation is outdated for the newer variants (i.e. Sudoku and the new dark variants we added)
- Same goes for `variantslogic/suit.go`, this is just replica of the code that we have in `src/suits.go` already (but with less error checks)

Possible fix:
Since we still want some logic calculating efficiency for these stats pages, we should probably get rid of the `variantslogic` directory entirely (only duplicated or outdated code) and re-implement that logic just in the `src/variant.go` file.

kesslermaximilian:
Fixes #2913.

This removes the variantslogic directory entirely: It contained mostly duplicate code, the rest was outdated.
The relevant logic for calculating efficiencies is now available as part of `src/variant.go`, which seems like the proper place.

I compared the newly calculated the efficiencies to the old ones, the only differences were:
- Sudoku 4 Suits. The new values are the correct ones now, I compared them to the ones calculated by the client now and they match (and also apart from that they seem correct to me)
- Up Or Down + Dark Suit (6 Suits). Here as well, the newly calculated ones match the client and should be the correct ones
- Any variant that involves a dark reversed suit. This is because the old code was bugged there already: When parsing the `suits.json` file, the code in the `variantslogic` directory did not add any reversed suits. When looking up such suits, instead of reporting an error, it returned a default suit and therefore did not account for suits as 'Black Reversed' to be dark and thus have only 5 cards instead of 10. This resulted in wrong deck sizes and thus efficiencies. Since the new implementation uses the (already-existing, it's unclear to me why it wasn't done so already) function to calculate deck sizes from `src/variant.go`, this is fixed as well now, becasue there the suits are initialized properly.

So up to my knowledge, all these changes should be correct.

Side effect: The `variantslogic` code pre-calculated all the efficiencies and essentially cached them so that they only need to be looked up on request. The current implementation (this PR) does not do so.
If you think this is necessary (it feels to me that this is not very dramatic computation on the side, but who knows, not an expert in go), then I can also implement that. In that case, I would suggest to just add another attribute to the `Variant` class, that seems like the easiest place.


kesslermaximilian:
Also, this PR deletes more lines than it adds, so it's gotta be good ;)

PR #2915, opened 2023-12-16
Add assertions regarding retrieval of cards from globals.deck

kesslermaximilian:
Since we believe that
https://github.com/Hanabi-Live/hanabi-live/issues/2894 is related to getCardOrStackBase returning 'wrong' cards, i.e. cards that are not marked 'isStackBase' where we would expect so, we introduce additional checks upon calling this method to hopefully catch errors sooner.

Additionally, we check that `globals.deck` never gets too long.


Zamiell:
very good pr thank you

PR #2916, opened 2023-12-18
SLR: Clean up Sudoku code

kesslermaximilian:
This also fixes a bug: In Sudoku 4-suit, the minimum should be taken with variant.stackSize, not with just the default number of suits.

### DIFF
--- a/packages/game/src/rules/variants/sudoku.ts
+++ b/packages/game/src/rules/variants/sudoku.ts
@@ -70,8 +70,8 @@ export function sudokuWalkUpAll(
     if (allDiscardedSet.has(currentRank)) {
       // We hit a new dead rank.
       for (const writeRank of eRange(lastDead + 1, currentRank)) {
-        maxScoresForEachStartingValueOfSuit[writeRank - 1] =
-          currentRank - writeRank;
+        const maxScore = currentRank - writeRank;
+        maxScoresForEachStartingValueOfSuit[writeRank - 1] = maxScore;
       }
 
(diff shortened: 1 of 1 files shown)


packages/game/src/rules/variants/sudoku.ts
```diff
@@ -71,8 +70,8 @@ export function sudokuWalkUpAll(
     if (allDiscardedSet.has(currentRank)) {
       // We hit a new dead rank.
       for (const writeRank of eRange(lastDead + 1, currentRank)) {
-        const maxScore = currentRank - writeRank;
-        maxScoresForEachStartingValueOfSuit[writeRank - 1] = maxScore;
+        maxScoresForEachStartingValueOfSuit[writeRank - 1] =
```
Zamiell, line 73: why are you removing my SLR change here?

kesslermaximilian: Linting suggested it, but I don't have a real preference on it.

Zamiell: what lint suggested it?

kesslermaximilian: Oh I guess it's not linting rules, it's my IDE itself, I'll revert it and re-configure.

Zamiell: sounds like a terrible feature, its optimizing for the "smallest number of expressions", which is usually the exact opposite of "easiest code to read"

PR #2918, opened 2023-12-20
fix parsing colors in hypos

kesslermaximilian:
This fixes the bug introduced in ad384ac during refactoring, where the `startsWith` method was replaced with an exact match.


Zamiell:
thx

PR #2928, opened 2024-01-25
[Throw it in a Hole] Fix sound & stack bugs

NZadeh:
As I was trying to tackle https://github.com/Hanabi-Live/hanabi-live/issues/2875, I thought this low-hanging fruit would be a good way to start.  Does this commit look okay to you?

EDIT: This is the error that should no longer be reachable: https://github.com/Hanabi-Live/hanabi-live/blob/main/packages/client/src/game/ui/reactive/getSoundType.ts#L208

We're now computing the number of blind "plays" successfully. Previously, it was set to 0 for TIIAH bombs.

### DIFF
--- a/packages/client/src/game/ui/HanabiCard.ts
+++ b/packages/client/src/game/ui/HanabiCard.ts
@@ -570,13 +570,13 @@ export class HanabiCard extends Konva.Group implements NodeWithTooltip, UICard {
   }
 
   /**
-   * getBareNameRank is used to compute the card's "bare name", which is used to render the correct
+   * Used to compute the card's "bare name", which is used to render the correct
    * image (e.g. "card-${suit}-${bareNameRank}"). The "card" image when nothing is on the stacks has
    * the special name "card-${suit}-${CARD_IMAGE_STACK_BASE_RANK_NAME}".
    */
   getBareNameRank(rankToShow: Rank | null): string {
-    // If a non-null rank is provided, it should take the highest precedence
-    // (e.g. a card note, including on a stack base in "Throw it in a Hole").
+    // If a non-null rank is provided, it should take the highest precedence.
+    // (We want players to make notes on the stack bases in "Throw it in a Hole" variants.)
     if (rankToShow !== null) {
       return rankToShow.toString();
     }
--- a/packages/client/src/game/ui/HanabiCard.ts
+++ b/packages/client/src/game/ui/HanabiCard.ts
@@ -570,15 +570,17 @@ export class HanabiCard extends Konva.Group implements NodeWithTooltip, UICard {
   }
 
   getBareNameRank(rankToShow: Rank | null): string {
+    // If a non-null rank is provided, it should take the highest precedence.
+    // E.g. a note on a card (including a stack base in "Throw it in a Hole").
+    if (rankToShow !== null) {
+      return rankToShow.toString();
+    }
+
     if (this.isStackBase) {
       return CARD_IMAGE_STACK_BASE_RANK_NAME;
     }
 
-    if (rankToShow === null) {
-      return CARD_IMAGE_UNKNOWN_CARD_RANK_NAME;
-    }
-
-    return rankToShow.toString();
+    return CARD_IMAGE_UNKNOWN_CARD_RANK_NAME;
   }
 
   // --------------
--- a/packages/client/src/game/ui/HanabiCard.ts
+++ b/packages/client/src/game/ui/HanabiCard.ts
@@ -569,9 +569,14 @@ export class HanabiCard extends Konva.Group implements NodeWithTooltip, UICard {
     return `card-${bareNameSuit}-${bareNameRank}`;
   }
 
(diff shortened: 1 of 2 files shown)


NZadeh:
@Zamiell PTAL!

NZadeh:
I also folded the stack fixes into this change.

image

I'm... sort of holding on for dear life in terms of figuring out how to actually contribute to this repo, and this whole forking and merging model is super confusing rn.  So hopefully everything is "correct-enough" process-wise and sorry I couldn't do it better.

packages/client/src/game/ui/HanabiCard.ts
```diff
@@ -570,15 +570,17 @@ export class HanabiCard extends Konva.Group implements NodeWithTooltip, UICard {
   }
 
   getBareNameRank(rankToShow: Rank | null): string {
-    if (this.isStackBase) {
-      return CARD_IMAGE_STACK_BASE_RANK_NAME;
+    // If a non-null rank is provided, it should take the highest precedence.
```
Zamiell, line 573: what is the rank of a stack base? can we just delete the "if (stackbase)" part?

Zamiell:
> I'm... sort of holding on for dear life in terms of figuring out how to actually contribute to this repo, and this whole forking and merging model is super confusing rn

if you aren't familiar with github then you can dm me and i can help, otherwise im sure you can just youtube any old "how to do a PR" explainer video, its fairly straightforward

packages/client/src/game/ui/HanabiCard.ts
NZadeh: `getBareNameRank` is used to compute the card's "bare name" which is used to render the correct image.  Something of the form `card-${bareNameSuit}-${bareNameRank}`.

So you might see `card-Teal-1` for the teal-1s in people's hands or if a teal 1 is on the stacks. If there's nothing on the stacks (the sort of "hollowed-out placeholder"), that has the name `card-Teal-${CARD_IMAGE_STACK_BASE_RANK_NAME}`.

The idea here is to allow users to (once again) override the stack-base name by having `rankToShow` apply first. But we _definitely_ still want to back off to CARD_IMAGE_STACK_BASE_RANK_NAME so that we show the stack placeholder in case there is no note or the stack is empty.

packages/game/src/reducers/statsReducer.ts
```diff
@@ -216,8 +217,17 @@ function statsReducerFunction(
...
-  if (action.type !== "play") {
+function isBlindPlay(
+  action: GameAction,
+  gameState: GameState,
+  variant: Variant,
+): boolean {
+  // In "Throw it in a Hole" variants, bombs should appear as successful plays.
+  const possiblePlay =
+    action.type == "play" ||
+    (variant.throwItInAHole && action.type == "discard" && action.failed);
```
NZadeh, line 228: whoops, sorry, I botched the `===` on these. Is this easy for you to fix?

Zamiell: you can just fix it yourself by pushing another commit to the fork

packages/client/src/game/ui/HanabiCard.ts
Zamiell: then you should then copy that whole github comment into an actual code comment that fully explains what the code is doing

NZadeh: happy to do that. (it's not like I wrote the code, btw... xD I just spent a while debugging it to figure out what it's doing. good idea to let future folks know what it does too...!)

will try to do this in the next commit.

packages/game/src/reducers/statsReducer.ts
NZadeh: ack. is there a way for me to run the lints myself before the website does it? to figure out what's broken?

packages/client/src/game/ui/HanabiCard.ts
Zamiell: right, the idea is that we want to leave the code in a better state than when we started

packages/game/src/reducers/statsReducer.ts
Zamiell: yeah just run lint.sh

Zamiell: but you should see the errors in your editor (vscode) immediately as you are coding, otherwise you are going to be having a real shitty time, dm me on discord and i can help you set it up

NZadeh: thanks for the help over discord! also fixed in the latest commit.

packages/client/src/game/ui/HanabiCard.ts
NZadeh: love it! (done)

NZadeh:
Alright should be fixed, and ./lint.sh passes locally for me! Thanks for the help! :D

Zamiell:
good PR thanks

PR #2963, opened 2024-09-03
Join pre-game with stored join password

progiv:
When pre-game url is opened - send join request with stored join password.

I play hanabi mostly with my friends on password
protected tables. Every time I open the pre-game link, It tells me: "That is not the correct password for the game". I didn't provide any password, yet. But the game gives me an error message.

This PR fixes my use case, and doesn't break games without passords Tested this way (U1, U2 - users):
1. U1: Create game without password.
2. U2: Join the pre-game link
3. U1, U2 leave table
4. U1: Create password protected game
5. U2: Join the pre-game link. Notice the wrong password message
6. U2: Enters password, successfully joins the table
6. U1, U2 leave game
7. U1: Create password proteced game. (Same password)
8. U2: Join the pre-game link. **No errors**
9. U1, U2 leave game
10. U1 crate game without password
11. U2 Join the pre-game link. No errors (the stored password is still sent because of this PR)

### DIFF
--- a/packages/client/src/lobby/url.ts
+++ b/packages/client/src/lobby/url.ts
@@ -61,6 +61,8 @@ export function parseAndGoto(data: ServerCommandWelcomeData): void {
     if (tableIDString !== undefined) {
       const tableID = parseIntSafe(tableIDString);
       if (tableID !== undefined && tableID > 0) {
+        // Here we unconditionally send password from cookies. If the game has no password, the
+        // server will ignore it.
         const password = localStorage.getItem("joinTablePassword");
         globals.conn!.send("tableJoin", {
           tableID,
--- a/packages/client/src/lobby/url.ts
+++ b/packages/client/src/lobby/url.ts
@@ -61,8 +61,10 @@ export function parseAndGoto(data: ServerCommandWelcomeData): void {
     if (tableIDString !== undefined) {
       const tableID = parseIntSafe(tableIDString);
       if (tableID !== undefined && tableID > 0) {
+        const password = localStorage.getItem("joinTablePassword");
         globals.conn!.send("tableJoin", {
           tableID,
+          password
         });
         return;
       }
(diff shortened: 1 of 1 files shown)


Zamiell:
what happens when you try to reconnect to a game that doesn't have a password, but you send a password for the game anyway

progiv:
> what happens when you try to connect to a game that doesn't have a password, but you send a password for the game anyway

It just works. The client joins the table (steps 11,12 in the first message) The password is sent, but it is never checked. See [server code](https://github.com/Hanabi-Live/hanabi-live/blob/d7e7581ca445124ce9477d3b6c843b84d0e93224/server/src/command_table_join.go#L55) 
```go
if t.PasswordHash != "" && !d.BypassPassword {
    ...
```
If the game has no password, first part of the `if` becomes false, and this branch is not executed


> CI format failed

This _prettier_ is a nasty thing. I managed to do the PR without installing npm on my computer by fixing docker-compose file locally, but now I have to install the whole dev environment because the _prettier_ doesn't show what's wrong(((

Zamiell:
yeah sorry its nasty for tiny one-off pull requests, but in the general case auto-formatters like prettier are so good that they will change your life forever

Zamiell:
> It just works. The client joins the table (steps 11,12 in the first message) The password is sent, but it is never checked. See [server code](https://github.com/Hanabi-Live/hanabi-live/blob/d7e7581ca445124ce9477d3b6c843b84d0e93224/server/src/command_table_join.go#L55)

then you should put that as a comment in the PR so it will be clear for people reading the code in the future

progiv:
>  in the general case auto-formatters like prettier are so good that they will change your life forever

I do love auto-formatter in general. But since it already knows what's wrong in my code, I expect them to show the diff. It seems like an industry standard, this particular _prettier_ doesn't follow.

Anyway, thank you for keeping this repo clean and readable!

Zamiell:
thanks for the pr

Issue #2937, opened 2024-03-29
Variant Request: Inverted Suit


jakestiles:
Cards in the Inverted Suit play when discarded and discard when played.

**Example gameplay**
Example A: Alice clues Bob fuchsia (or whatever color gets assigned to this suit) touching one card.  Bob discards it which causes it to successfully play as the Inverted 1.
Example B: Alice discards her chop which is unclued. It was the Inverted 4 so it plays unsuccessfully and generates a strike.
Example C: All 1s are on the stacks. Alice has an unknown 2 in her hands. She doesn't know whether she should play it to play it or discard it to play it.

**Art**
The art for these cards could perhaps be made by turning the number values (and/or the symbols) upside down and/or backwards.

**Coding exception**
A small extra coding exception has to be made for this suit. Normally you may not discard at 8 clues. But cards can be played at any time. Thus if a player plays an Inverted card at 8 clues, it should be discarded and should fail to generate a 9th clue.

**Name and implementation**
I'm fine with the suit being renamed. Maybe Silly or Opposite or Disobedient. Or just named after whatever color it gets assigned.
This might work better as a suit modifier (like Reversed) than a suit of its own (like Null).

Issue #1720, opened 2020-09-01
New Variant Dump


pianoblook:
Had a bit of fun tonight brainstorming some new Variant ideas:

**1. Ivory**: White, but the cards are only touched by the number one above their actual value (so the 1 is only touched by 2, 2 touched by 3...and the 5 wraps back to being touched by 1)

**2. Void**: Null, but they only successfully play by being discarded (who's ready for a Null & Void game?)

**3A. Thorn**: Causes a strike whenever this is clued [could also work as a modifier: e.g. Thorny White]

**3B. Gold**: Causes a strike whenever one of these is discarded [also could just be a modifier]

> A less punishing take on Thorn & Gold would be that you just lose a clue instead

**4. Polkadot**: Turn order reverses whenever one of these cards (successfully?) plays

**5. Parasite**: Gets touched by any clue that would otherwise only touch a single card. Can't be touched otherwise.

**6. Ash**: Scores points as they are discarded in the correct order. Playing any Ash card, or discarding one out of order, causes a strike. Already-scored duplicates can be discarded normally.

jakestiles:
Interested in Gold (discard to strike)
Interested in Gold v2 (discard yields no clue?
Very interested in Polkadot. 

Zam would you accept PR’s for any/all of these if I did the work?

Zamiell:
honestly discard to strike doesn't sound very fun, your hand just gets super locked with gold cards.
im not against adding variants but i dont want to add unfun variants just for the sake of adding more variants

jakestiles:
Ok. Would you accept PRs for Gold v2 or Polkadot?

jakestiles:
I’m not a score chaser nor a full time coder, so I can assure you’d I’m only asking about the variants I think I’d have fun playing. Not just trying to bloat the list.

Zamiell:
how does the end game work in polkadot

jakestiles:
When end game is triggered in a normal game, you get PlayerCount more turns. Same for Polka Dot.  So not everyone might get a “final” turn. Polka Dot distribution may lead to unwinnable games similar to black suits.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

padiwik:
Shouldn't it be consistent with Contrarian? (currently a 2-turn end game) Or does the balancing need to be different for a polka dot suit vs. a contrarian character?

Zamiell:
gold v2 sounds pretty borderline not very interesting

[...]

PR #2968, opened 2024-09-10
Add inverted suit

IceSelkie:
Inverted Suit as proposed in #2937 (and also one of the many suits proposed in #1720 (Ash)).

The inverted suit should play when you try discard, and discard when you try to play.

This seems to work fairly well for the client side view, however a few issues have been found when I was testing it with some friends.

- [x] Bottom Deck Play option doesn't work yet. It would need to be possible to discard from bottom deck to be able to play an inverted card. 
- [x] Reading from the database will double the inversion, causing the replay to be invalid and cut off part way through.
- [x] NEW: Merge/rebase onto the new main, since relevant code changed.

Todo (beyond fixing above)
- Maybe change the animation to be like misplay for discards, and a misplay but in the other direction for successful plays.

Will not do (see Zamiell's comment below):
- Interactions with determental characters may cause issues.

Where I am hooking in to make the play/discard swap might not be the best place. If someone who knows the code better knows a better place to put this in, do let me know.


Zamiell:
i wouldn't worry about interactions with detrimental characters, they will probably be deleted in the long term, as they are too much of a maintenance burden

IceSelkie:
Original post edited to reflect that detrimental characters won't be considered, and that a merge/rebase needs to be done to make it mergeable again.

ricardodd2:
> burden

I would be kinda sad for it to be completely delated. Maybe make it as an option only for no var/make it a variant

IceSelkie:
Merged to match main; fixed to pass CI/lint; made bottom-deck-plays also allow discards for inverted games.

https://github.com/IceSelkie/hanabi-live/actions/runs/11310399814

The main functionality is working, so this can probably be merged now.

A few improvements could be made in the future, and I don't know if I'll have much time to investigate these:
- [ ] Inverted plays/discards could use a new better animation/sound so they look/sound distinct from other plays were it not inverted
- [ ] Bottom-Deck Discard is currently allowed on all games, where the server then checks if the game has inverted or not, giving a warning if it shouldn't be allowed. (Previously, the dragged bottom-deck card would float above discard pile without discarding, playing, returning, or giving a warning)

Zamiell:
i think it should be some other color instead of black?
black denotes 1 of each

IceSelkie:
Oh right, I meant to change that. I'll can go add a new Color of "Inverted". (Though it does works fine now and look how I want; I'm not sure if the word "black" or "K" show up anywhere at the moment, and you can't get any black in the same game as inverted, since I only added "Inverted (5 suits)" and "Inverted (6 suits)" as of now)

Zamiell:
by the way, why not inverted 3 suits and 4 suits?

IceSelkie:
I believe this PR should be ready now.

Zamiell:
nice thank you, ill see if i can find some time this weekend to review again, IIRC from before it only needed some small changes

Zamiell:
Sorry I completely forgot about this, I'll see if I can review it in the next day or two

PR #2976, opened 2024-11-15
Add Sticky Brown Basic Variants & Create a Variant Walkthrough Contributing.md

Fioralor:
Adds Variants where Colored Non-Brown Suits have been modified to also be able to be Clued by the Brown Color Clue to add an additional challenge to the Brown Suit


Fioralor:
It's all green Finally :). Have added a Create a new Variant walkthrough as requested. Feel free to edit Contributing.md as you see fit. I have only created 1 Variant Group so have probably missed some things. As far as the Variant is concerned, I made decisions in my new Suit ID's that I thought did not conflict with any existing patterns I found for already existing suit id's. For example the Modified Red Suit ID SRN. I came to with S for Sticky R for the Clue Color Red and N for the Clue Color Brown. There were some other patterns I thought maybe these suits would fit into and am open to changing the suit id's if you prefer a different pattern. Trying to think ahead to when the new variant scheme without variant numbers but with strictly suit id's takes effect and trying to keep things as simple as possible :).

Zamiell:
can you remove dark brown from the PR? new variants are supposed to be trial, meaning only with the base suits

packages/game/src/json/suits.json
```diff
@@ -573,5 +573,40 @@
     "displayName": "Teal",
     "pip": "hanabiAltWhite",
     "clueColors": ["Green", "Blue"]
+  },
+  {
+    "name": "Red N",
```
Zamiell, line 578: what does N stand for?

Zamiell:
is there an existing issue for the Sticky Brown variant?

Fioralor:
What is meant by existing issue?

Zamiell:
a github issue proposing how the variant should work and so on

Fioralor:
It looks like contributing.md has been updated already to add the create a variant walkthrough. I have updated the pr contributing.md to match it. I'm not sure if I should just rename the pr again or keep it since part of this pr's history was helping to add the create a variant walkthrough. I'm fine with either option.

Fioralor:
I had some irl stuff come up for the past few weeks. I meant to close my own pr with a comment. Self Closing this pr for the variant idea being too similar to already existing variants.

PR #2986, opened 2025-02-03
Add support for specifying max number of players via "/create-table" URL (?maxPlayers= or ?numPlayers=)

NZadeh:
Some players on the website have expressed interest in this.

Pointers on how to test this are appreciated. (I suspect it works just from browsing similar code... but a little suspicious about how maxPlayers is specified in the WebSocket call).


NZadeh:
Can't figure out correct formatting on the Web IDE... :'(

Does someone want to take this over and format it properly?

Zamiell:
dont use the web ide

PR #2996, opened 2025-06-08
Add tab focusing after clicking on notification

GregTag:
Just small feature


Zamiell:
can you make it pass ci

GregTag:
Ok, this time I've launched linter locally. I should have done this initially. It should be fine now.

Issue #2997, opened 2025-06-20
spectators in lobby bug


Zamiell:
if you:
- join a non-started game lobby
- and then return to lobby (without exiting)
- and then spec a game with other spectators
- and the go back to lobby and go back to your table

then the wrong specs will be listed

PR #2999, opened 2025-06-29
Fix: update spectators list when reattending

### LINKED ISSUE #2836: spectate live game with other spectators, then resume lobby table, spectators list is bugged

GregTag:
Closes #2836 and #2997


server/src/command_table_spectate.go
GregTag, line 1: Removed second call of `t.NotifySpectators()`, first call at line 123

Zamiell:
oh i guess i opened duplicate issues lol

PR #3002, opened 2025-07-19
fix: chat history when remaking a table

GregTag:
When remaking table pregame previous chat history wasn't sent.
Bug reproduction:
* Create a table
* Start the game
* Terminate game and remake the table
* See no previous messages (for any player)


Zamiell:
thx!

PR #3006, opened 2025-07-25
Add White Reversed & Rainbow variant

mg6-hanabi:
From time to time someone comes from BGA or elsewhere asking if this official variant can be played here, so I thought it would be nice to finally have it.

The only questionable thing is if it should be named "White Reversed & Rainbow" or maybe a more natural "Rainbow & White Reversed". The name I've chosen seems less confusing as the latter might suggest both Rainbow & White are reversed.

If this gets a green light, I'm in favor of adding the remaining variants including, but not limited to:
Reversed & Rainbow
Reversed & Pink
Reversed & White
Reversed & Brown
Reversed & Omni
Reversed & Null
Reversed & Muddy Rainbow
Reversed & Light Pink
Reversed & Prism

Rainbow Reversed & Rainbow? hmm, maybe, maybe not; there is Rainbow & Dark Rainbow after all...
Rainbow Reversed & Pink
...
Rainbow Reversed & Prism

White Reversed & Rainbow... done
White Reversed & Pink
White Reversed & White? same as Rainbow Reversed & Rainbow
...
White Reversed & Prism

Prism Reversed & Rainbow
...
Prism Reversed & Light Pink
Prism Reversed & Prism? quirky

And maybe even these special double reversed:
Reversed & Rainbow Reversed
Reversed & White Reversed
Reversed & Prism Reversed
Rainbow Reversed & White Reversed
Rainbow Reversed & Prism Reversed
White Reversed & Prism Reversed


mg6-hanabi:
The most compelling reason to have this specific variant is that from time to time people come here and ask about it.
Here is a recent message I've found: https://discord.com/channels/140016142600241152/140016142600241152/1352856614129897564
And one of the reasons they do is probably that it the the most advenced official variant in the box: https://boardgamegeek.com/boardgame/290357/hanabi-deluxe-what-a-show

At minimum I think it would be good to support this specific combination, it's the most popular among bga gamers and from time to time you might see someone come and ask about this variant.

Zamiell:
fine

PR #3008, opened 2025-07-26
Feat: join game / spectate pregame navigation button

GregTag:
Adds button to navigation panel in pregame screen.
h1
h2
h3

### DIFF
--- a/packages/client/src/lobby/nav.ts
+++ b/packages/client/src/lobby/nav.ts
@@ -6,6 +6,7 @@ import * as tooltips from "../tooltips";
 import * as createGame from "./createGame";
 import * as history from "./history";
 import * as pregame from "./pregame";
+import * as tablesDraw from "./tablesDraw";
 import * as watchReplay from "./watchReplay";
 
 export function init(): void {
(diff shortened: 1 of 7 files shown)


GregTag:
> from the screenshots, it looks like it says "Join" for both joining and spectating. maybe it should instead say "Spectate" for the spectate case?

I wanted to make it consistent with hotkeys as Alt + J for both options. Should then Spectate have different hotkey and if so, which one?

Zamiell:
> I wanted to make it consistent with hotkeys as Alt + J for both options.

i think the hotkeys should probably be a secondary consideration.

> Should then Spectate have different hotkey and if so, which one?

that's completely up to you, maybe try to find some letter that isn't being used for anything else?

packages/client/src/lobby/nav.ts
```diff
@@ -138,6 +139,29 @@ export function init(): void {
...
+        return;
+      }
+
+      if (
+        table.spectators.some(
+          (spectator) => spectator.name === globals.username,
+        )
+      ) {
+        // We are a spectator. We can join table without unattending it.
+        tablesDraw.tableJoin(table);
```
mg6-hanabi, line 156: Shouldn't there be a call to tableLeave in case you have joined another table and spectating this one? I often have a table open and then go back to lobby and join another table as spectator.

GregTag: You are right, I hadn't thought of that possibility. I'll fix it

packages/client/src/lobby/tablesDraw.ts
```diff
@@ -262,10 +262,6 @@ export function tableSpectate(
   table: ServerCommandTableData,
   shadowingPlayerIndex = -1,
 ): void {
-  if (globals.currentScreen !== Screen.Lobby) {
```
Zamiell, line 265: why are we removing this

GregTag: It allows to reuse `tableSpectate` and `tableJoin` functions and avoid copying code functions at `nav.ts`. This should not affect the rest of the function code nor the behavior of these functions in other parts, since they were always called on the Lobby screen.

GregTag:
> it looks like the PR has a bugfix to the password input field? that seems unrelated to this feature though?

Yes it is. I came across this during the testing. Should I put this in a new PR?

packages/client/src/lobby/tablesDraw.ts
Zamiell: but shouldn't it just be changed to `if (globals.currentScreen !== Screen.Lobby && globals.currentScreen !== Screen.PreGame) {`?

Zamiell:
yes, please put unrelated changes in a separate PR, otherwise it is madness

packages/client/src/lobby/nav.ts
Zamiell: is this one resolved now?

GregTag: Yes, I've tested it with different options, should work now

Zamiell:
great work thank you!

PR #3034, opened 2025-11-02
fix: disable popup warning for inverted suits

### LINKED ISSUE #3029: Popup warning when "playing" inverted suit
**Describe the bug**
When you drag and inverted card to the stacks ("play" it, to make it discard and get a hint back) if the card is known to not be playable, you will get a popup warning that the card cannot be playable. Since it's inverted, the card is going to be discarded, not played, so this behavior is nonsensical.

**To Reproduce**

1. Start an Inverted game
2. Have i1 played on the stacks
3. Fully clue an i1 in a players hand
4. Have that player drag the card to the stacks, to make it discard
5. A popup will appear warning that the card is not playable

**Expected behavior**
Unplayable warnings should happen either when discarding a known inverted card, or never for inverted cards.
Alternatively, could just have always considering inverted cards as "playable" could be a workaround if that's possible.

vEnhance:
"Playing" an inverted suit actually discards it instead.

Closes #3029

### DIFF
--- a/packages/game/src/rules/card.ts
+++ b/packages/game/src/rules/card.ts
@@ -174,6 +174,12 @@ export function isCardPotentiallyPlayable(
   return card.possibleCards.some((possibleCard) => {
     const [suitIndex, rank] = possibleCard;
 
+    // Always consider inverted cards to be "playable".
+    const suit = variant.suits[suitIndex];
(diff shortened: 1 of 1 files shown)


Zamiell:
thanks

PR #3036, opened 2025-11-04
Fix broken hanabi reference links

tanmay2004:
Hi, I found some links to the hanabi reference in the codebase which were broken and am fixing them here


Zamiell:
thanks

PR #3039, opened 2025-11-15
Refactor !seed table naming: add/increment numeric suffix reliably

CardSwim8:
Changes:
- Introduce incrementSeedName(name string) to centralize !seed naming logic.
- Detect trailing digits using a precompiled regexp (`^(.*?)(\d+)$`).
- Increment numeric suffix if present, else append "1".

Rationale:
!seed games follow a similar naming progression to normal tables.
Ensuring deterministic numeric increments improves ease of use for players who restart or reproduce seed-based setups.

Verification:
- gofmt -w and go vet ./... return clean
- go test ./... (server/src) passes, including new cases for !seed handling
- install_dependencies.sh rebuilds client and server without errors
- Local server startup validated up to the database connection step

### DIFF
--- a/server/src/command_table_restart.go
+++ b/server/src/command_table_restart.go
@@ -9,6 +9,24 @@ import (
 	"github.com/Hanabi-Live/hanabi-live/logger"
 )
 
+var trailingNum = regexp.MustCompile(`^(.*?)(\d+)$`)
+
+// incrementSeedName takes a !seed-style name and either:
+// - increments trailing digits, if they exist ("!seed abc1" -> "!seed abc2")
+// - or appends "1" if there are no trailing digits ("!seed abc" -> "!seed abc1")
+func incrementSeedName(name string) string {
+	if m := trailingNum.FindStringSubmatch(name); len(m) == 3 {
+		// m[0] = full match
+		// m[1] = base (everything before the digits)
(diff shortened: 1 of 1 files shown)


Zamiell:
did you create a table and test to ensure that it works correctly in both contexts?
e.g. "foo1" and "foo"

alexvornsand:
Yes, I wrote a test file that included the following examples. I didn't include it in the PR since I didn't see any other test files, but I'd be happy to add it to the PR. These are tested against the `incrementSeedName` function, which is only invoked when there is a `!seed` prefix.

```
        {
            name: "no trailing number, simple seed name",
            in:   "!seed abc",
            want: "!seed abc1",
        },
        {
            name: "existing trailing 1 gets incremented",
            in:   "!seed abc1",
            want: "!seed abc2",
        },
        {
            name: "multi-digit number increments correctly",
            in:   "!seed game42",
            want: "!seed game43",
        },
        {
            name: "no space before number still works",
            in:   "!seedthing9",
            want: "!seedthing10",
        },
        {
            name: "name that already ends with digits but no seed prefix still works (helper is generic)",
            in:   "foobar7",
            want: "foobar8",
        },
        {
            name: "name with no digits at all",
            in:   "justname",
            want: "justname1",
        },
    }
```

PR #3060, opened 2026-01-28
Add a maroon border on cards with a 'ptd' note

JKTKops:
Hi! This was easy enough to implement that I figured I'd ask if you're interested in taking the PR. There is essentially no new maintenance overhead with this change and the diff is quite small.

In H-Group conventions, PTD is a highly contextual concept. When I was learning level 9 a while ago, I had a few games in a row where I received PTD but didn't want to discard just yet, and then on a later turn tried to protect my chop because I had forgotten. (Even after writing a "ptd" note, in one case.) In review, more experienced players told me that at level 9 I should be holding on to PTD until the card is clued. While I don't think expert players would benefit much from this change, I think it will help beginners and learners get familiar with the concept and remember when PTD is given much as we use the cm border to remember cards we should _not_ discard.

A significant side-benefit of this change is that other conventions are also played actively on the website. Several referential convention systems, including turbo, reactor, and ref-sieve, have conventional (i.e. not contextual) concepts of PTD on unclued cards. A border ala the finesse border is a very useful tool when playing with those systems to the point where I've seen some suggestions of implementing this with a client-side browser plugin. For this reason I've included 'ctd' as an abbreviation that triggers the border. I personally think that just for those convention systems, the change with _only_ 'ctd' as a trigger would still be a net positive UX.

I have no strong opinions about the specific color, if the one I've used happens to be an issue.

### DIFF
--- a/packages/client/src/game/ui/constants.ts
+++ b/packages/client/src/game/ui/constants.ts
@@ -40,8 +40,8 @@ export const CHOP_MOVE_COLOR = "#fffce6";
 
 export const FINESSE_COLOR = "aqua";
 
-/** A soft red, almost pink. */
-export const DISCARD_PERMISSION_COLOR = "#f07070";
+/** A dark shade of maroon. */
+export const DISCARD_PERMISSION_COLOR = "#812020";
 
 export const OFF_BLACK = "#0d0d0d";
 


Zamiell:
anyone else have any opinions on the color

Zamiell:
> Probably the only thing contentious about this change is that the other borders represent hard states of a card (except in the case of an ambiguous finesse) while PTD is more contextual.

That's true. Perhaps PTD should be represented by an image instead, similar to "trash" and "?". The former is the existing precedent for "I should discard this card". The latter is the existing precedent for "I have specific context about this card". As for what icon to choose, maybe a thumbs-up? Not really sure, just brainstorming.

JKTKops:
The reason I went for a border is because in other convention systems (where I think it might get more use as a feature) ptd/ctd is a hard state of a card. Even considering that it's a bit contentious for H-Group, I still think a border makes more sense.

If it were to be an image, I think a thumbs-down makes more sense than a thumbs-up. I'm also struggling to think of a good icon, but thumbs-down was my first idea when I thought of implementing a patch.

Gingy-120:
> I'm unsure whether this is the correct place to talk about this, but on a similar note, I think that similar to how gd draws a border for Gentleman's Discard, there should also be bd for Baton Discard. I'm not sure whether it should be the same finesse border or maybe something else, since cards don't play immediately from BDs, unlike GDs.

What I do is write [cm][b3] on the card if it's a Baton and [f][b3] if it's a Gentleman's.

PR #3063, opened 2026-02-01
Add Chop Move border to cards with the 'bd' note

Sir-Mastermind:
This PR gives cards with a `bd` note the Chop Move border.
The reason I chose the Chop Move border is that, unlike cards known by a (Layered) Gentleman's Discard, those known through a Baton Discard cannot be played immediately, so they shouldn't get the Finesse border.
This keeps in line with moves like 5NE where the partially known card gets a cm note.


Zamiell:
Thx

PR #3066, opened 2026-02-28
feat: Responsive UI resizing

anicolao:
This PR adds a debounced resize handler to automatically redraw the game board when the window is resized. It prevents having to perform a full page reload when changing the size of your browser.

### DIFF
--- a/packages/client/src/game/ui/drawUI.ts
+++ b/packages/client/src/game/ui/drawUI.ts
@@ -27,6 +27,7 @@ import { CardLayout } from "./CardLayout";
 import { ClueLog } from "./ClueLog";
 import { ColorButton } from "./ColorButton";
 import { Deck } from "./Deck";
+import { Elements } from "./Elements";
 import { FullActionLog } from "./FullActionLog";
 import { HanabiCard } from "./HanabiCard";
 import { MultiFitText } from "./MultiFitText";
@@ -60,7 +61,6 @@ import * as replay from "./replay";
 import * as stats from "./stats";
 import * as timer from "./timer";
 import { toggleZen } from "./zen";
-import { Elements } from "./Elements";
 
 interface Values {
   x: number;


Zamiell:
furthermore, we already decided against doing this, as it was too buggy.
see:

1. e0aefe84 - "Resize canvas when the window changes size (#1634)" — The original PR that added automatic resizing (Aug 16, 2020)
  2. 93fab7ca - "Revert "Resize canvas when the window changes size (#1634)"" — First revert (Aug 17, 2020)
  3. 7a684ec9 - "Revert "Revert "Resize canvas when the window changes size (#1634)"" " — Re-applied the feature (Aug 17, 2020)
  4. 01fb1b0c - "Revert "Revert "Revert "Resize canvas when the window changes size (#1634)""" " — Final revert, with the message: "stay with old behavior until bugs are fixed" (Aug 17, 2020)

flackr:
We played a bunch of 4 player games with this today across several devices (Mac, Linux, iPad) and browsers (Chrome, Firefox), encouraging everyone to resize regularly throughout the game. There were only 2 real issues:
1. Every resize replays the loading animations as if the page is refreshed
2. If you are viewing a clue in a replay, the clue indicator is lost when you resize until you move forward / backward to that turn.

While (1) is moderately annoying it's still far better than having to reload the entire game to resize the UI. The issue in (2) seems likely that most of the time you wouldn't notice in that as soon as you change turns the indicators are back.

TLDR, I think this approach is working much better than the previous PRs, and offers a much more seamless experience when you need to change your browser window size.

flackr:
We discovered one more minor bug. When viewing the history, if someone takes an action, it replays the move in the history that you're looking at as if it is happening again.

E.g.
Alice clues 1,
Cathy is looking at the 1's clue in the replay history,
Bob does anything else, it looks like Bob is giving 1's clue.

anicolao:
Thanks for clarifying, that makes sense.

I’ve updated this so replay-history viewing still gets the sound notification when a new live action happens, while keeping the no-animation behavior so the historical move you’re looking at doesn’t appear to replay.

Could you retest that same scenario and confirm this now matches what you want?

Zamiell:
thank you for this high quality PR!

PR #3068, opened 2026-02-28
feat: add Nix development environment support

anicolao:
This PR adds a `flake.nix` file to provide a reproducible development environment using Nix.

### DIFF
--- a/.envrc.nix.example
+++ b/.envrc.nix.example
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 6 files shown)


anicolao:
Sorry, I didn't provide install instructions for direnv, but if they are allowed they do require direnv to be in the spellcheck. Take a look at the changes proposed to install.md and if it's OK to list nix as a supported environment, then this PR is ready. Also included bun for better performance on nix-darwin.

PR #3072, opened 2026-03-02
Add identity token infrastructure, /token chat command, and identity lookup APIs

alexvornsand:
## Summary
This PR adds lightweight identity-token infrastructure so authenticated users can prove account ownership externally.

It introduces:
- a per-user identity token persisted in Postgres
- a private `/token` chat command that generates or retrieves an active token
- authenticated API endpoints to retrieve/regenerate a token
- a public unauthenticated API endpoint to resolve `username` from token

## Problem
We need a simple server-native way for users to obtain a short-lived secret token and then use that token externally to verify identity.

### Chat UX
- Added `/token` command.
- `/token` is private-response only and does **not** appear in public chat output/history.
- `/token` behavior:
  - returns existing active token if still valid
  - otherwise generates a new token and returns it
- Any extra arguments return usage guidance.

### Token Rules
- Exactly one active token per user (`PRIMARY KEY (user_id)` + upsert behavior).
- Regeneration replaces prior token.
- Token TTL is 24 hours.
- Token is generated from cryptographically secure random bytes and base64url-encoded.

### APIs
- `GET /api/v1/identity/token` (authenticated)
  - returns existing active token for logged-in user
- `POST /api/v1/identity/token` (authenticated)
  - regenerates and returns token for logged-in user
- `GET /api/v1/identity/:token` (public, unauthenticated)
  - returns username for valid, non-expired token

## Database
Added new table in schema install path:
- `user_identity_tokens`
  - `user_id` (PK, FK -> `users.id`)
  - `token` (UNIQUE)
  - `expires_at`
  - `datetime_created`
  - `datetime_updated`

## Files Changed
- `install/database_schema.sql`
- `server/src/models.go`
- `server/src/models_user_identity_tokens.go` (new)
- `server/src/identity_token.go` (new)
- `server/src/api_misc.go`
- `server/src/api_identity.go` (new)
- `server/src/chat_command.go`
- `server/src/chat_identity_token.go` (new)
- `server/src/command_chat.go`

## Verification
Ran CI-equivalent checks locally:
- `bash ./packages/client/build_client.sh`
- `bash ./server/build_server.sh`
- `bash ./packages/client/lint.sh`
- `bash ./packages/data/lint.sh`
- `bash ./packages/game/lint.sh`
- `bash ./packages/scripts/lint.sh`
- `bash ./packages/server/lint.sh`
- `npm run test`
- `npx prettier --log-level=warn --check .`
- `bash check_shellcheck.sh`
- `bash check_spelling.sh`
- `bash check_templates.sh`
- `npm run check-variant-files`
- `npm run lint-package-json`

All passed.

## Notes
- PR intentionally stays in narrow feature scope (no Dockerfile/tooling/config drift).

### DIFF
--- a/packages/server/jest.config.cjs
+++ b/packages/server/jest.config.cjs
@@ -1,0 +1,30 @@
+/* eslint-disable import-x/no-extraneous-dependencies */
+const fs = require("node:fs");
+const path = require("node:path");
+const jsoncParser = require("jsonc-parser");
+const { pathsToModuleNameMapper } = require("ts-jest");
+
+// Read and parse the compiler options from the "tsconfig.json" file.
+const REPO_ROOT = path.resolve(__dirname, "..", "..");
+const MONOREPO_TS_CONFIG = path.join(REPO_ROOT, "tsconfig.monorepo.json");
+if (!fs.existsSync(MONOREPO_TS_CONFIG)) {
+  throw new Error(`The "${MONOREPO_TS_CONFIG}" file does not exist.`);
+}
+
+const tsconfigString = fs.readFileSync(MONOREPO_TS_CONFIG).toString().trim();
+const tsconfig = jsoncParser.parse(tsconfigString);
+
+/** @type {import("ts-jest/dist/types").JestConfigWithTsJest} */
+const config = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+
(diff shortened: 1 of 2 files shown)


Zamiell:
shouldn't we be using the same hashing algorithm that passwords use though?

Zamiell:
thanks, let's try it out!

PR #3075, opened 2026-03-14
fix: add missing migration script for user_identity_tokens table

alexvornsand:
## Summary

- PR #3072 deployed the identity token feature but did not include a migration script to create the `user_identity_tokens` table in the production database
- The `database_schema.sql` changes can't be run against prod (it's a destructive full-rebuild), and no standalone migration was provided
- As a result, `/token` fails immediately with the generic error modal because every DB call hits a non-existent table
- This PR adds `server/migrate_user_identity_tokens_add_hashing.sql`, following the same convention as `server/migrate_sound_move.sql` from #3030, to create the table with the correct final schema


Zamiell:
this pr is not necessary because the "database_schema.sql" file is idempotent right?

Zamiell:
i just pasted the contents into a database shell, so it should be good to go now

PR #3076, opened 2026-03-15
docs: reuse identity token infra in alpha staging design

anicolao:
Updates the alpha-ui staging doc to reuse the new identity-token infrastructure as the auth base, instead of describing a separate token system.


Zamiell:
seems fine

Issue #2855, opened 2023-07-27
Server Crash: Concurrent read/write on map


kesslermaximilian:
James posted this crash log on Discord:

[message.txt](https://github.com/Hanabi-Live/hanabi-live/files/12179609/message.txt)

kesslermaximilian:
Probably this is a bug introduced in commit e74d6bdf in the the `ActiveSpectators` method in `table.go:433`.

Zamiell:
@Krycke

Krycke:
Yeah, the record of spectators on the table hold a list of all spectators that have been specific since the table was created. The spectator may or may not spectate right now. 

The only trust worthy record of if a spectator is spectating actively right now is the record in tables that we get from the function `tables.GetTablesUserSpectating(sp.UserID)`. 

So now from the point of view of the table, we need to go through the list of all spectators, and see if any of them are spectating a table right now, and if they take appear to be this one.

This function in it self should not be the problem here, i have a hard time reading the stack trace on my phone, but i assume that there is a deadlock. 

When i fixed this function, i was also confused about what we are locking and when, this function it self does not need the tables to be locked, if they are changed during checking it wouldn't matter, but this function uses functions that are might or might not need a lock.

The problem with finding dead locks in the stack trace is that it's not this thread that might be the problem, it's the thread that has claimed the lock the is suck somewhere that probably is the problem. But that one is not seen in the trace.

When i was trying to investigate the dead locks before i remember that the following could occur: `thread 1` wants to first acquire `lock 1` and while having that lock, it is trying to acquire `lock 2`. The second lock is already kept by a `thread 2`, so `thread 1` gets stuck. At the same time a `thread 3` is trying to acquire `lock 1` and eventually times out. That's why we find the stack trace from the dead lock on `thread 3` but i think the problem is actually that `thread 2` won't let go of `lock 2` or that `thread 1` is trying to acquire a second lock before releasing the first one.

Either way i think all locks need to be checked in all functions, because something is strange with them.

kesslermaximilian:
I think the function here should definitely have the lock for the tables struct when being executed. I can see why we wouldn't care if the spectators list changes while being read, but (presumably) these reads are not atomic, so stuff might go wrong if another thread writes these at the same time.
The error message from the trace explicitly says 'concurrent read and write', so it could very much be that the stack trace we see is the thread reading from the tables struct, while some other thread (hopefully holding the lock for that) is currently writing it because some spectator joined.

[...]

PR #3077, opened 2026-04-02
Fix bug #2855 of hanabi-live

panfs:
I fix issue https://github.com/Hanabi-Live/hanabi-live/issues/2885. 
Now notes parser can read and handle cases with a protected note and a common note after '|'.
For example:
1. Player A notes one of his card(called X) [cm].
2. Player A notes card X, appended 'y2'.

What's happen now:
**Before:**
Card X only appears as y2, without chop move block surrounded.
**After:**
Card X correctly appears as y2 with chop move block surrounded.

Other cases, for example, [f] and 'r24', [cm] and 'bg35', handles correctly as well.


Zamiell:
nice fix ><

PR #3078, opened 2026-04-03
Add a Shift + alt + right-click keyboard shortcut for adding ptd notes.

sjdrodge:
Now that we have ptd borders, let's have a keyboard shortcut for putting them on cards.

### DIFF
--- a/packages/client/src/game/ui/clickNotes.ts
+++ b/packages/client/src/game/ui/clickNotes.ts
@@ -52,7 +52,7 @@ export function clickRightCheckAddNote(
     return;
   }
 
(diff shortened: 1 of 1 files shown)


packages/client/src/game/ui/clickNotes.ts
```diff
@@ -52,6 +52,14 @@ export function clickRightCheckAddNote(
     return;
   }
 
+  // Shift + alt + right-lick is a "ptd" note. (This is a common abbreviation for "this card has
```
Zamiell: right lick my butt

PR #3081, opened 2026-04-04
Restore the note parsing logic.

panfs:
Revert the note parsing logic to the previous judgment logic, i.e., only using the most recent content to determine, so that player intentions are shown while preserving note history. Regarding issue 2885, I believe their suggestion is incompatible with the habits of most current players, so no changes will be made. If you want to display both, for example, chopMove board and y2, you should use [cm][y2], not other variants like [cm]|y2 or [cm] y2.

### DIFF
--- a/docs/alpha-ui-staging.md
+++ b/docs/alpha-ui-staging.md
@@ -65,7 +65,7 @@ Use three pieces:
 1. `alpha preview UI` (GitHub Pages, per PR)
 1. `prod server` (single source of truth for games)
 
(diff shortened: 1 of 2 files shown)


Zamiell:
this PR has no changes

DarthGandalf:
I think you need to merge master into this branch, then it'll show correctly. This PR contains commits which do the change (which are in master already) AND commits which revert it. So with the old base it indeed shows empty in this UI

Zamiell:
i just reverted the commit manually

Issue #3082, opened 2026-04-06
Sudoku: cards incorrectly (not) greyed out


samuelweiler:
**Describe the bug**
In Sudoku mode, sometimes unplayable cards are not correctly marked as such.

**To Reproduce**

1. [Replay at turn 29](https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2FHanabi-Live%2Fhanabi-live%2Ftree%2Fmain)
2. Purple is not started yet and must start with a 1.  Both purple 2's are in the trash (oops). 
3. Newly-drawn purple 3 is correctly greyed out
4. Held purple 4 and 5 are incorrectly *not* greyed out
5. The bug first appears at [turn 21](https://hanab.live/replay/1809544#21), just after the second purple 2 is tossed, when the 5 should be greyed.

**Expected behavior**
Held purple 4 and 5 should be greyed out, too.

**Screenshots**

<img width="1186" height="678" alt="Image" src="https://github.com/user-attachments/assets/faa2765f-5008-4544-93ac-800d22c58c2d" />

**Additional context**
In a very quick scan other other teams' sub-25 scores, both on this deal and others, I have not yet found another example of this error.

Zamiell:
@kesslermaximilian Can you fix this?

samuelweiler:
@kesslermaximilian, thanks for the variant; we've been enjoying it.

PR #3087, opened 2026-04-13
Fix: fix Sudoku card fade reactivity for cross-suit updates(issue #3082)

panfs:
**Why bug occurs and how to test this fix**
- In previous logic, the appearance(faded or not) of color X is changed only when one X-color card is played or discarded. This logic is acceptable in most varients, but it may encounter rare edge cases shown below.
- We assume that there is a 4-player sudoku-varient game, and red starts with 3, yellow starts with 2, and blue starts with 5. There is a p5 held by some player. Now green and purple can begin with either 1 or 4.
- Because of some mistakes, all p2 are discarded. Now p5 still can be played if purple begins with 4.
- In another step, green begins with 4. According to sudoku's rule, purple can ONLY begins with 1. And because ALL p2 are discarded. p5 cannot be played anymore.
- After the analysis, we can find that due to sudoku's specific rule, the previous logic cannot handle this edge cases.
- To test this fix, you can simulate this edge case shown above.

**Commit description**
- `Sudoku: recompute cardStatus for ALL suits/ranks after play/discard to reflect cross-suit stack-start constraints.`
- `Handle scrubbed play actions by deriving suitIndex from the updated deck when available so playStackStarts/directions update reliably.`
- `Client: compute per-card status subscription using the same identity source as rendering (morphed → deck identity → cardIdentities), so trash/critical fades update immediately without requiring extra actions.`


Zamiell:
can you add a test

packages/game/src/reducers/gameReducer.ts
```diff
@@ -530,16 +541,33 @@ function gameReducerFunction(
...
-      gameState.cardStatus[action.suitIndex][rank] = getCardStatus(
-        action.suitIndex,
-        rank,
-        gameState.deck,
-        gameState.playStacks,
-        gameState.playStackDirections,
-        gameState.playStackStarts,
-        variant,
-      );
+    if (variant.sudoku) {
```
Zamiell: eh i dont like the variant specific code.
can you just simplify it to look through the whole deck, regardless of variant?
the perf wont matter afaik

packages/game/src/reducers/gameReducer.ts
```diff
@@ -487,40 +488,50 @@ function gameReducerFunction(
...
-    );
-
-    gameState.playStackStarts[action.suitIndex] = getStackStartRank(
-      playStack,
-      gameState.deck,
-    );
+  if (action.type === "play" && variant.sudoku) {
+    const suitIndex =
+      action.suitIndex === -1
+        ? (gameState.deck[action.order]?.suitIndex ?? null)
```
Zamiell: use undefined instead of null please

packages/game/src/reducers/gameReducer.ts
panfs: I've revised it.

packages/game/src/reducers/gameReducer.ts
```diff
@@ -530,16 +541,19 @@ function gameReducerFunction(
...
-      gameState.cardStatus[action.suitIndex][rank] = getCardStatus(
-        action.suitIndex,
-        rank,
-        gameState.deck,
-        gameState.playStacks,
-        gameState.playStackDirections,
-        gameState.playStackStarts,
-        variant,
-      );
+    for (const i of variant.suits.keys()) {
```
Zamiell, line 547: can you put a code comment saying:
```
For a no-variant game, we would only have to iterate over cards of the same suit to check for non-playbility. But other variants become more complicated, so we iterate over the entire deck to be safe.
```

panfs: alright.

Zamiell:
thanks. did you playtest this change on your local server to ensure that it works as expected?

panfs:
Yes, it works as expected.

**Before y4 is played:**

截图 2026-04-14 00-56-43

---

**After y4 is played:**

截图 2026-04-14 00-57-28

Zamiell:
excellent, thank you!

PR #3089, opened 2026-04-15
fix: pre-fetch pregame stats in commandTableJoin to complete deadlok fix

Pimak:
The deadlock fixes in 11a3f29b and 52796122 addressed tableRestart and tableCreate, which previously held BOTH tables.Lock + t.Lock while issuing DB queries (GetUserNumGames + UserStats.Get). Under DB pool exhaustion this trio (tables.Lock + t.Lock + DB) cycled into a deadlock that go-deadlock reliably caught and aborted the process for.

The user-initiated path (commandTableJoin -> tableJoin) was left with the same lock-then-DB footprint: tableJoin acquires tables.Lock while t.Lock is already held by getTableAndLock, then issues the same two DB roundtrips. Concurrent joins under load therefore retain the original deadlock surface.

Move the pre-fetch up into commandTableJoin so the DB roundtrips happen under t.Lock only (tables RLock from getTable is released before this point). This matches the pattern the prior fixes established for internal callers (which pre-fetch even earlier under no locks at all and pass the result via d.PregameStats).

The legacy DB path inside tableJoin is preserved as a defensive fallback for any caller that forgets to populate d.PregameStats, and is documented as such.

### DIFF
--- a/server/src/command_table_join.go
+++ b/server/src/command_table_join.go
@@ -70,13 +70,45 @@ func commandTableJoin(ctx context.Context, s *Session, d *CommandData) {
 		return
 	}
 
+	// Pre-fetch the joining player's pregame stats here, BEFORE tableJoin acquires tables.Lock.
+	// At this point only t.Lock is held (acquired by getTableAndLock above; the brief tables RLock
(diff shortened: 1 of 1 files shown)


Zamiell:
thanks

PR #3093, opened 2026-05-19
add /split, /splitauto and /sa command

Teiem:
When playing Hanabi, sometimes your lobby fills up to a point where it makes sense to split the table into two new groups. Those new commands, `/split [username1] [username2]...`, `/splitauto` and its alias `/sa` intend to make this a bit easier.

- `/split` creates a new lobby with the users listed
- `/splitauto` and `/sa` do the same as `/split`, except that the users are automatically determined by total games count.

It would be ideal for it to be possible for spectators to be included in those commands, but that could allow for forced participation in games, which is undesirable. Maybe a button in chat like for suggested variants might be a solution for that in the future.

While I am a developer, I do not know Go, therefore all of this code has been written with the help of copilot. I have tested it and it seems to work, but I cant tell how well the code is written, or if it is correct besides the general logic.

### DIFF
--- a/docs/chat-commands.md
+++ b/docs/chat-commands.md
@@ -49,20 +49,20 @@ If you need general help with the website, then read the [features page](feature
 
 ## Pre-game Commands (table-owner-only)
 
-| Command | Description |
-| --- | --- |
-| `/impostor` | Randomly tells one of the players they are an impostor and the others they are crew-mates. |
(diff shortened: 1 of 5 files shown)


Zamiell:
can you make it pass CI though

Teiem:
Should be fine now? I formatted `chat-commands.md` with prettier.

Zamiell:
thanks

PR #3097, opened 2026-06-13
Change arrow color of cm cards with HGroup feature

Teiem:
Did you ever get the focus of a clue wrong because the chop was chop moved? This PR changes the arrow color of chop moved cards to be light grey (`#737373`), between the color of untouched and touched cards. 

This is only enabled if the "Enable [H-Group](https://github.com/Hanabi-Live/hanabi-live/blob/main/docs/features.md#h-group-conventions) features" setting is enabled.

Note: when adding a cm note to a card while an arrow is visible, the arrow will not update and stay white. This should usually not be a problem since a clue that touches a card can not also chop move it. Having the arrow reappear will then show the correct color (e.g. go back and forth in history)

### DIFF
--- a/packages/client/src/game/ui/arrows.ts
+++ b/packages/client/src/game/ui/arrows.ts
@@ -26,6 +26,29 @@ import { drawPip } from "./drawPip";
 import { getCardOrStackBase } from "./getCardOrStackBase";
 import * as konvaHelpers from "./konvaHelpers";
 
+function getClueArrowColor(
+  element: Konva.Node | null,
+  preview: boolean,
+): string {
+  if (!(element instanceof HanabiCard)) {
+    return ARROW_COLOR.DEFAULT;
+  }
+
+  if (
+    element.state.numPositiveClues >= 2
+    || (element.state.numPositiveClues >= 1 && preview)
+  ) {
+    return ARROW_COLOR.RETOUCHED;
+  }
+
+  // if H-Group features are enabled, use a different color for cards that have been chop-moved
+  if (globals.lobby.settings.hyphenatedConventions && element.note.chopMoved) {
+    return ARROW_COLOR.CM_TOUCHED;
+  }
+
+  return ARROW_COLOR.DEFAULT;
+}
+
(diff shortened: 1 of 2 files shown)


packages/client/src/game/ui/constants.ts
```diff
@@ -29,6 +29,8 @@ export const ARROW_COLOR = {
   /** Dark gray */
   RETOUCHED: "#737373",
 
+  CM_TOUCHED: "#B9B9B9",
```
Zamiell, line 33: write a comment that explains what the color is

packages/client/src/game/ui/arrows.ts
```diff
@@ -26,6 +26,29 @@ import { drawPip } from "./drawPip";
...
+function getClueArrowColor(
+  element: Konva.Node | null,
+  preview: boolean,
+): string {
+  if (!(element instanceof HanabiCard)) {
+    return ARROW_COLOR.DEFAULT;
+  }
+
+  if (
+    element.state.numPositiveClues >= 2
```
Zamiell, line 41: explain this line

Teiem: This is kept from the preexisting code:

```ts
    // This is a clue arrow.
    const color =
      element instanceof HanabiCard
      && (element.state.numPositiveClues >= 2
        || (element.state.numPositiveClues >= 1 && preview))
        ? ARROW_COLOR.RETOUCHED // Cards that are re-clued use a different color.
        : ARROW_COLOR.DEFAULT; // Freshly touched cards use the default color.
```
        
  I have added the following comment:
  
  ```ts
  // The arrow is drawn after the clue is given, so by this time `numPositiveClues` having a value
  // of 2 means it had one clue already. During preview the clue is not applied yet, so if it has at
  // least one positive clue on it already, it will be retouched by the next clue.
  ```

Zamiell:
this is probably a good change, but note that since the colors are so similar, in reality it is going to be pretty hard for players to notice the tiny difference in a real game

PR #3098, opened 2026-06-18
Allow omitting 0 before . for time in /si command

Teiem:
This PR will allow the `/startin [time]` and `/si [time]` commands to accept a time that starts with `.`, e.g. `.5` for `0.5`.


Zamiell:
did you test that the change works as expected?

Teiem:
I did with some common inputs like `.5` `0.5` and `1`. Is there another accepted number format that I forgot, esp. one where a leading 0 would make a difference?

PR #3102, opened 2026-07-18
Replace broken link to pips-reference script

3geek14:
The current link leads to a 404. This should fix it, but I haven't built it locally to test. Worst-case, it's still a 404.

### DIFF
--- a/public/standalone/pips-reference.html
+++ b/public/standalone/pips-reference.html
@@ -122,6 +122,6 @@
       <div class="loading">Loading pips...</div>
     </div>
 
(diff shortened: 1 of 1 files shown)


Zamiell:
I don't think this will work actually, can you test to confirm?

3geek14:
Apparently not, as I've been unable to get the database working on my machine to finish the installation instructions. :( I might keep trying to get it to work.

Zamiell:
yeah this didn't fix it, see this commit which seems to fix it https://github.com/Hanabi-Live/hanabi-live/commit/246d0374e4b5bcd22da056a3024533e27048637f

